PR #5, opened 2019-04-22
Protocol changes for R4.1

marmarek:
- Increase data chunk size to 64k.
- Increase allowed length of requested qrexec service name - MSG_TRIGGER_SERVICE message. 

Fixes https://github.com/QubesOS/qubes-issues/issues/4909

### DIFF
--- a/agent/qrexec-agent-data.c
+++ b/agent/qrexec-agent-data.c
@@ -40,6 +40,8 @@
 
 #define VCHAN_BUFFER_SIZE 65536
 
+#define QREXEC_DATA_MIN_VERSION QREXEC_PROTOCOL_V2
+
 static volatile int child_exited;
 static volatile int stdio_socket_requested;
 int stdout_msg_type = MSG_DATA_STDOUT;
@@ -105,7 +107,7 @@ int handle_handshake(libvchan_t *ctrl)
 
     actual_version = info.version < QREXEC_PROTOCOL_VERSION ? info.version : QREXEC_PROTOCOL_VERSION;
 
-    if (actual_version != QREXEC_PROTOCOL_VERSION) {
+    if (actual_version < QREXEC_DATA_MIN_VERSION) {
         fprintf(stderr, "Incompatible agent protocol version (remote %d, local %d)\n", info.version, QREXEC_PROTOCOL_VERSION);
         return -1;
     }
@@ -153,17 +155,18 @@ void send_exit_code(libvchan_t *data_vchan, int status)
  *  1 - some data processed, call it again when buffer space and more data
  *      available
  */
-int handle_input(libvchan_t *vchan, int fd, int msg_type)
+int handle_input(libvchan_t *vchan, int fd, int msg_type, int data_protocol_version)
 {
-    char buf[MAX_DATA_CHUNK];
+    const size_t max_len = max_data_chunk_size(data_protocol_version);
+    char buf[max_len];
     ssize_t len;
     struct msg_header hdr;
 
     hdr.type = msg_type;
     while (libvchan_buffer_space(vchan) > (int)sizeof(struct msg_header)) {
         len = libvchan_buffer_space(vchan)-sizeof(struct msg_header);
-        if ((size_t)len > sizeof(buf))
-            len = sizeof(buf);
+        if ((size_t)len > max_len)
+            len = max_len;
         len = read(fd, buf, len);
         if (len < 0) {
             if (errno == EAGAIN || errno == EWOULDBLOCK)
@@ -202,10 +205,11 @@ int handle_input(libvchan_t *vchan, int fd, int msg_type)
  */
 
 int handle_remote_data(libvchan_t *data_vchan, int stdin_fd, int *status,
-        struct buffer *stdin_buf)
+        struct buffer *stdin_buf, int data_protocol_version)
 {
     struct msg_header hdr;
-    char buf[MAX_DATA_CHUNK];
+    const size_t max_len = max_data_chunk_size(data_protocol_version);
+    char buf[max_len];
 
     /* do not receive any data if we have something already buffered */
     switch (flush_client_data(stdin_fd, stdin_buf)) {
@@ -221,9 +225,9 @@ int handle_remote_data(libvchan_t *data_vchan, int stdin_fd, int *status,
     while (libvchan_data_ready(data_vchan) > 0) {
         if (libvchan_recv(data_vchan, &hdr, sizeof(hdr)) < 0)
             return -1;
-        if (hdr.len > MAX_DATA_CHUNK) {
+        if (hdr.len > max_len) {
             fprintf(stderr, "Too big data chunk received: %" PRIu32 " > %zu\n",
-                    hdr.len, MAX_DATA_CHUNK);
+                    hdr.len, max_len);
             return -1;
         }
         if (!read_vchan_all(data_vchan, buf, hdr.len))
@@ -289,7 +293,8 @@ int handle_remote_data(libvchan_t *data_vchan, int stdin_fd, int *status,
 }
 
 int process_child_io(libvchan_t *data_vchan,
-        int stdin_fd, int stdout_fd, int stderr_fd)
+        int stdin_fd, int stdout_fd, int stderr_fd,
+        int data_protocol_version)
 {
     fd_set rdset, wrset;
     int vchan_fd;
@@ -408,7 +413,10 @@ int process_child_io(libvchan_t *data_vchan,
         }
 
         /* handle_remote_data will check if any data is available */
-        switch (handle_remote_data(data_vchan, stdin_fd, &remote_process_status, &stdin_buf)) {
+        switch (handle_remote_data(data_vchan, stdin_fd,
+                    &remote_process_status,
+                    &stdin_buf,
+                    data_protocol_version)) {
             case -1:
                 handle_vchan_error("read");
                 break;
@@ -432,7 +440,8 @@ int process_child_io(libvchan_t *data_vchan,
                 break;
         }
         if (stdout_fd >= 0 && FD_ISSET(stdout_fd, &rdset)) {
-            switch (handle_input(data_vchan, stdout_fd, stdout_msg_type)) {
+            switch (handle_input(data_vchan, stdout_fd, stdout_msg_type,
+                        data_protocol_version)) {
                 case -1:
                     handle_vchan_error("send");
                     break;
@@ -442,7 +451,8 @@ int process_child_io(libvchan_t *data_vchan,
             }
         }
         if (stderr_fd >= 0 && FD_ISSET(stderr_fd, &rdset)) {
-            switch (handle_input(data_vchan, stderr_fd, MSG_DATA_STDERR)) {
+            switch (handle_input(data_vchan, stderr_fd, MSG_DATA_STDERR,
+                        data_protocol_version)) {
                 case -1:
                     handle_vchan_error("send");
                     break;
@@ -506,6 +516,7 @@ int handle_new_process_common(int type, int connect_domain, int connect_port,
     libvchan_t *data_vchan;
     int exit_code = 0;
     pid_t pid;
+    int data_protocol_version;
 
     if (type != MSG_SERVICE_CONNECT) {
         assert(cmdline != NULL);
@@ -527,7 +538,7 @@ int handle_new_process_common(int type, int connect_domain, int connect_port,
         fprintf(stderr, "Data vchan connection failed\n");
         exit(1);
     }
-    handle_handshake(data_vchan);
+    data_protocol_version = handle_handshake(data_vchan);
 
     prepare_child_env();
     /* TODO: use setresuid to allow child process to actually send the signal? */
@@ -540,13 +551,17 @@ int handle_new_process_common(int type, int connect_domain, int connect_port,
             do_fork_exec(cmdline, &pid, &stdin_fd, &stdout_fd, &stderr_fd);
             fprintf(stderr, "executed %s pid %d\n", cmdline, pid);
             child_process_pid = pid;
-            exit_code = process_child_io(data_vchan, stdin_fd, stdout_fd, stderr_fd);
+            exit_code = process_child_io(
+                    data_vchan, stdin_fd, stdout_fd, stderr_fd,
+                    data_protocol_version);
             fprintf(stderr, "pid %d exited with %d\n", pid, exit_code);
             break;
         case MSG_SERVICE_CONNECT:
             child_process_pid = 0;
             stdout_msg_type = MSG_DATA_STDIN;
-            exit_code = process_child_io(data_vchan, stdin_fd, stdout_fd, stderr_fd);
+            exit_code = process_child_io(
+                    data_vchan, stdin_fd, stdout_fd, stderr_fd,
+                    data_protocol_version);
             break;
     }
     libvchan_close(data_vchan);
--- a/agent/qrexec-agent-data.c
+++ b/agent/qrexec-agent-data.c
@@ -24,6 +24,7 @@
 #include <signal.h>
 #include <unistd.h>
 #include <errno.h>
+#include <inttypes.h>
 #include <string.h>
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -155,13 +156,13 @@ void send_exit_code(libvchan_t *data_vchan, int status)
 int handle_input(libvchan_t *vchan, int fd, int msg_type)
 {
     char buf[MAX_DATA_CHUNK];
-    int len;
+    ssize_t len;
     struct msg_header hdr;
 
     hdr.type = msg_type;
     while (libvchan_buffer_space(vchan) > (int)sizeof(struct msg_header)) {
         len = libvchan_buffer_space(vchan)-sizeof(struct msg_header);
-        if (len > (int)sizeof(buf))
+        if ((size_t)len > sizeof(buf))
             len = sizeof(buf);
         len = read(fd, buf, len);
         if (len < 0) {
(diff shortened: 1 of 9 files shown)


marmarek:
This PR could use some extra pair of eyes on review, both on design level and implementation level.
cc @jpouellet @HW42

jpouellet:
I've started to look, but I need to refresh my brain of the surrounding subsystems first (and the changes since I've last looked) to have a deep enough understanding and sufficient context in order to provide a meaningful review. It might be a little while.

agent/qrexec-agent-data.c
```diff
@@ -152,17 +156,18 @@ void send_exit_code(libvchan_t *data_vchan, int status)
  *  1 - some data processed, call it again when buffer space and more data
  *      available
  */
-int handle_input(libvchan_t *vchan, int fd, int msg_type)
+int handle_input(libvchan_t *vchan, int fd, int msg_type, int data_protocol_version)
 {
-    char buf[MAX_DATA_CHUNK];
-    int len;
+    const size_t max_len = data_protocol_version < 3 ? MAX_DATA_CHUNK_V2 : MAX_DATA_CHUNK_V3;
+    char buf[max_len];
```
HW42: I have seen that some code bases try to avoid allocating bigger chunks on the stack, and often avoid dynamic stack allocation altogether. Since the size has a fixed maximum it's probably ok as is.

agent/qrexec-agent-data.c
```diff
@@ -152,17 +156,18 @@ void send_exit_code(libvchan_t *data_vchan, int status)
...
+    ssize_t len;
     struct msg_header hdr;
 
     hdr.type = msg_type;
     while (libvchan_buffer_space(vchan) > (int)sizeof(struct msg_header)) {
         len = libvchan_buffer_space(vchan)-sizeof(struct msg_header);
-        if (len > (int)sizeof(buf))
-            len = sizeof(buf);
+        if (len > (int)max_len)
```
HW42: cast to int is wrong. Probably we want `(size_t)len > max_len`? (The mixing of `size_t` and `ssize_t` in functions like `read` is really annoying)

agent/qrexec-agent-data.c
```diff
@@ -220,9 +226,9 @@ int handle_remote_data(libvchan_t *data_vchan, int stdin_fd, int *status,
     while (libvchan_data_ready(data_vchan) > 0) {
         if (libvchan_recv(data_vchan, &hdr, sizeof(hdr)) < 0)
             return -1;
-        if (hdr.len > MAX_DATA_CHUNK) {
-            fprintf(stderr, "Too big data chunk received: %d > %d\n",
-                    hdr.len, MAX_DATA_CHUNK);
+        if (hdr.len > max_len) {
+            fprintf(stderr, "Too big data chunk received: %d > %lu\n",
```
HW42: `hdr.len` is `uint32_t` and `max_len` is `size_t` so `%" PRIu32 "` and `%zu` are the correct formats, I think.

agent/qrexec-agent.c
```diff
@@ -868,28 +868,51 @@ int fill_fds_for_select(fd_set * rdset, fd_set * wrset)
...
-        if (libvchan_send(ctrl_vchan, &params, sizeof(params)) < 0)
-            handle_vchan_error("write params");
-    }
-    if (ret <= 0) {
-        close(client_fd);
+    if (!read_all(client_fd, &hdr, sizeof(hdr)))
+        goto error;
+    if (hdr.type != MSG_TRIGGER_SERVICE3 ||
+            hdr.len < sizeof(params) ||
+            hdr.len > MAX_SERVICE_NAME_LEN) {
```
HW42: I think you want `hdr.len > sizeof(params) + MAX_SERVICE_LEN` here. This would also match the code in `sanitize_message_from_agent`.

agent/qrexec-agent.c
```diff
@@ -868,28 +868,51 @@ int fill_fds_for_select(fd_set * rdset, fd_set * wrset)
...
+    if (!read_all(client_fd, &params, sizeof(params)))
+        goto error;
+    command_len = hdr.len - sizeof(params);
+    command = malloc(command_len);
+    if (!command)
+        goto error;
+    if (!read_all(client_fd, command, command_len))
+        goto error;
+    /* check '\0' at the end */
+    if (command[command_len-1])
```
HW42: `if (command[command_len-1] != '\0')` and then drop redundant comment.

agent/qrexec-agent.c
```diff
@@ -868,28 +868,51 @@ int fill_fds_for_select(fd_set * rdset, fd_set * wrset)
...
+        goto error;
+
+    snprintf(params.request_id.ident, sizeof(params.request_id), "SOCKET%d", client_fd);
+    if (libvchan_send(ctrl_vchan, &hdr, sizeof(hdr)) < 0)
+        handle_vchan_error("write hdr");
+    if (libvchan_send(ctrl_vchan, &params, sizeof(params)) < 0)
+        handle_vchan_error("write params");
+    if (libvchan_send(ctrl_vchan, command, command_len) < 0)
+        handle_vchan_error("write params");
```
HW42: `"write command"`

agent/qrexec-client-vm.c
```diff
@@ -132,10 +135,16 @@ int main(int argc, char **argv)
         start_local_process = 1;
     }
 
+    service_name = argv[optind + 1];
+
+    service_name_len_plus_one = strlen(service_name) + 1;
```
HW42: All other code here seems to use length as a synonym for size (as supposed to length == strlen(x) == size - 1 as seen in some other code). So this should be just `service_name_len`.

daemon/qrexec-client.c
```diff
@@ -40,6 +40,10 @@ int replace_chars_stderr = 0;
 
 #define VCHAN_BUFFER_SIZE 65536
 
+#if MAX_DATA_CHUNK_V3 < MAX_DATA_CHUNK_V2
+#error "MAX_DATA_CHUNK_V3 must not be smaller than MAX_DATA_CHUNK_V2"
+#endif
```
HW42: Do wee need this check twice? Maybe move it into the header?

daemon/qrexec-client.c
```diff
@@ -70,7 +76,9 @@ static int handle_agent_handshake(libvchan_t *vchan, int remote_send_first)
                 return -1;
             }
 
-            if (info.version != QREXEC_PROTOCOL_VERSION) {
+            data_protocol_version = info.version < QREXEC_PROTOCOL_VERSION ?
+                                    info.version : QREXEC_PROTOCOL_VERSION;
+            if (data_protocol_version < 2) {
```
HW42: Formally hardcoded constants are against the style guide. I personally think this is fine but maybe consider to use something like `QREXEC_MIN_VERSION` or so.

There are a few other cases of hardcoded 2s and 3s. I won't tag them all.

daemon/qrexec-client.c
```diff
@@ -280,16 +288,21 @@ static void send_exit_code(libvchan_t *vchan, int status)
     }
 }
 
-static void handle_input(libvchan_t *vchan)
+static void handle_input(libvchan_t *vchan, int data_protocol_version)
 {
-    char buf[MAX_DATA_CHUNK];
+    char buf[MAX_DATA_CHUNK_V3];
```
HW42: Why do you allocate here always for V3?

daemon/qrexec-client.c
```diff
@@ -280,16 +288,21 @@ static void send_exit_code(libvchan_t *vchan, int status)
...
     int ret;
     size_t max_len;
     struct msg_header hdr;
 
-    max_len = libvchan_buffer_space(vchan)-sizeof(hdr);
+    max_len = libvchan_buffer_space(vchan);
+    if (max_len < sizeof(hdr))
+        return;
```
HW42: Don't we want some hard exit in this strange case?

daemon/qrexec-client.c
```diff
@@ -357,11 +370,12 @@ void do_replace_chars(char *buf, int len) {
...
 }
 
-static int handle_vchan_data(libvchan_t *vchan, struct buffer *stdin_buf)
+static int handle_vchan_data(libvchan_t *vchan, struct buffer *stdin_buf,
+        int data_protocol_version)
 {
     int status;
     struct msg_header hdr;
-    char buf[MAX_DATA_CHUNK];
+    char buf[MAX_DATA_CHUNK_V3];
```
HW42: Why always V3?

daemon/qrexec-daemon.c
```diff
@@ -553,7 +556,7 @@ static void handle_client_hello(int fd)
 /* handle data received from one of qrexec_client processes */
 static void handle_message_from_client(int fd)
 {
-    char buf[MAX_DATA_CHUNK];
+    char buf[1];
```
HW42: This change seems unrelated. Probably better moved to a separate commit.

daemon/qrexec-daemon.c
```diff
@@ -753,11 +747,32 @@ static void sanitize_message_from_agent(struct msg_header *untrusted_header)
...
+            if (protocol_version < 3) {
+                fprintf(stderr, "agent sent (new) MSG_TRIGGER_SERVICE3 "
+                        "although it uses protocol %d\n", protocol_version);
+                exit(1);
+            }
+            if (untrusted_header->len < sizeof(struct trigger_service_params3)) {
+                fprintf(stderr, "agent sent invalid MSG_TRIGGER_SERVICE3 packet\n");
+                exit(1);
+            }
+            if (untrusted_header->len-sizeof(struct trigger_service_params3)
```
HW42: I would put spaces around the minus.

daemon/qrexec-daemon.c
```diff
@@ -787,7 +806,52 @@ static void handle_message_from_agent(void)
...
+                    &params.request_id);
+            return;
+        case MSG_TRIGGER_SERVICE3:
+            service_name_len = hdr.len - sizeof(untrusted_params3);
+            untrusted_service_name = malloc(service_name_len);
+            if (!untrusted_service_name)
+                handle_vchan_error("malloc(service_name)");
+
+            if (libvchan_recv(vchan, &untrusted_params3, sizeof(untrusted_params3)) < 0)
+                handle_vchan_error("recv params");
```
HW42: maybe "recv params3" (same below)

libqrexec/qrexec.h
```diff
@@ -23,9 +23,15 @@
...
+#define QREXEC_PROTOCOL_VERSION 3
 #define MAX_FDS 256
-#define MAX_DATA_CHUNK 4096
+/* protocol version 2 */
+#define MAX_DATA_CHUNK_V2 4096
+/* protocol version 3+ */
+#define MAX_DATA_CHUNK_V3 65536
+
+/* large, but arbitrary */
+#define MAX_SERVICE_NAME_LEN 65536
```
HW42: Probably doesn't matter but why not a bit smaller such that the message will fit one chunk?

libqrexec/write-stdin.c
```diff
@@ -41,8 +41,8 @@ int flush_client_data(int fd, struct buffer *buffer)
         if (!len) {
             return WRITE_STDIN_OK;
         }
-        if (len > MAX_DATA_CHUNK)
-            len = MAX_DATA_CHUNK;
+        if (len > MAX_DATA_CHUNK_V3)
+            len = MAX_DATA_CHUNK_V3;
```
HW42: What about V2? I guess it's because it's used only on the agent (and you want to support only agent_version < daemon_version) but this looks like a rather generic helper function. So at least add a comment.

agent/qrexec-agent-data.c
```diff
@@ -39,6 +39,10 @@
 
 #define VCHAN_BUFFER_SIZE 65536
 
+#if MAX_DATA_CHUNK_V3 < MAX_DATA_CHUNK_V2
+#error "MAX_DATA_CHUNK_V3 must not be smaller than MAX_DATA_CHUNK_V2"
+#endif
```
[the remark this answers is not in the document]
marmarek: I rely on this in `handle_input()` - buffer size is based on `MAX_DATA_CHUNK_V3`, so its assume  to be enough for `MAX_DATA_CHUNK_V2` too. You found it already. But maybe indeed better don't do that at all.

agent/qrexec-agent-data.c
```diff
@@ -152,17 +156,18 @@ void send_exit_code(libvchan_t *data_vchan, int status)
  *  1 - some data processed, call it again when buffer space and more data
  *      available
  */
-int handle_input(libvchan_t *vchan, int fd, int msg_type)
+int handle_input(libvchan_t *vchan, int fd, int msg_type, int data_protocol_version)
 {
-    char buf[MAX_DATA_CHUNK];
-    int len;
+    const size_t max_len = data_protocol_version < 3 ? MAX_DATA_CHUNK_V2 : MAX_DATA_CHUNK_V3;
```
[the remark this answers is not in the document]
marmarek: Good idea.

daemon/qrexec-client.c
marmarek: I wanted to have it explicitly in `.c` file to guard against changing `.h` without considering all the files using it. Otherwise I'd add just a comment.

daemon/qrexec-client.c
marmarek: This isn't strange case. It may mean remote side haven't retrieved previous data yet.

libqrexec/qrexec.h
marmarek: Good idea.

daemon/qrexec-client.c
HW42: I personally would have added the assert in the header (maybe with a comment where/why this assumption is used/needed). Not only a comment such that the compiler catches a change even if someone didn't read the comment.

Anyway, your version is also reasonable.

As above maybe add a short comment.

agent/qrexec-agent-data.c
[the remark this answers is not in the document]
HW42: If you keep it, please add a short comment, like "used below to keep the code a bit easier" or something alike. Just to make it clear that why this constraint is introduced here.

daemon/qrexec-client.c
HW42: Oh, got confused about the meaning of `libvchan_buffer_space`.

HW42:
How is the version negotiation supposed to work for VM<->VM connections?

If I'm not mistaken we reach
https://github.com/QubesOS/qubes-core-qrexec/blob/d555fba6b23fd7103371cd54b14ed8ae108f0aad/agent/qrexec-agent-data.c#L109-L114
in `qrexec-client-vm` (via `handle_data_client`). So this would break with a domain with old qrexec, no?

HW42:
This seems to be fine, but just curious: Why do we respond with the highest supported protocol version, if we just decided that we need to emulate a older version:
https://github.com/QubesOS/qubes-core-qrexec/blob/d555fba6b23fd7103371cd54b14ed8ae108f0aad/daemon/qrexec-daemon.c#L221

libqrexec/write-stdin.c
marmarek: It doesn't really matter as this is communication with a local process, not remote side over vchan. I'll add a comment.

marmarek: Hmm, in fact the limit here doesn't make sense at all. If data chunk can't be received by the process as a whole, `write()` will return smaller number. But if it can, sending it in one call will be more performant. I'll remove the check.

marmarek:
> How is the version negotiation supposed to work for VM<->VM connections?

Indeed this is one place I forgot to update.

marmarek:
> This seems to be fine, but just curious: Why do we respond with the highest supported protocol version, if we just decided that we need to emulate a older version:

Mostly to have easier protocol specification and also care less about order of handshake messages. This reduces number of cases to think of to just "each side send its maximum supported version, the lower is used if supported by both sides".

Note that sending negotiated version as a response here may make an *illusion* that it will work with an agent before https://github.com/QubesOS/qubes-core-agent-linux/commit/1aeec025e00a1933c21c181a096a5580406ba8c8, but in practice it isn't true, as it doesn't apply to data channel. So, sticking to this protocol makes it easier to detect such problem.

agent/qrexec-agent-data.c
marmarek: I'll change that.
A simple test show that additionally heap buffer allocation is significantly faster.

```
[user@devel18 tmp]$ time ./alloc-test # stack

real	0m0.119s
user	0m0.118s
sys	0m0.001s
[user@devel18 tmp]$ time ./alloc-test 1 # heap

real	0m0.004s
user	0m0.002s
sys	0m0.001s
```

<details>

    #include <stdlib.h>
    #include <string.h>

    int alloc1(size_t s) {
        char buf[s];
        memset(buf, 0, sizeof(buf));
    }

    int alloc2(size_t s) {
        char *buf;
        buf = malloc(s);
        memset(buf, 0, sizeof(buf));
        free(buf);
    }

    int main(int argc, char **argv) {
        int i = 100000;

        if (argc == 1) {
            while (i--)
                alloc1(65536);
        } else {
            while (i--)
                alloc2(65536);
        }
    }

marmarek:
@HW42 I've force-pushed changes addressing your review.

marmarek:
@HW42 are the changes ok?

daemon/qrexec-client.c
```diff
@@ -280,18 +286,29 @@ static void send_exit_code(libvchan_t *vchan, int status)
...
+    max_len = libvchan_buffer_space(vchan);
+    if (max_len < sizeof(hdr))
+        return;
+    max_len -= sizeof(hdr);
+    if (max_len > buf_len)
+        max_len = buf_len;
     if (max_len == 0)
         return;
+
+    buf = malloc(buf_len);
```
HW42: Is there a reason for allocating `buf_len` instead of `max_len`? I would call `buf_len`, `data_chunk_size` or something and only allocate `max_len`.

HW42:
> @HW42 are the changes ok?

Mostly, see diff comment.

daemon/qrexec-client.c
marmarek: Makes sense. Force-pushed update.

marmarek:
@HW42 is it ok now?

PR #3, opened 2019-01-31
Working socket-based qrexec

DemiMarie:
This commit includes working socket-based qrexec.

Known limitations:

* Only dom0 →domU calls have been tested.
* dom0 part often hangs and must be killed manually

PR #6, opened 2019-08-09
Qrexec policy daemon

marmarta:
Rewritten qrexec-daemon to use policy daemon instead of running
policy-exec separately for each call. If daemon fails, falls back
to old solution. Also contains some tests.

fixes QubesOS/qubes-issues#5125

### DIFF
--- a/systemd/qubes-qrexec-policy-daemon.service
+++ b/systemd/qubes-qrexec-policy-daemon.service
@@ -1,0 +1,7 @@
+[Unit]
+Description=Qubes remote exec policy daemon
+After=qubesd.service
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/qrexec-policy-daemon
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ jobs:
        - pip install --quiet codecov
        - docker build -t qrexec-test ci
       env: DOCKER_RUN="docker run -v $TRAVIS_BUILD_DIR:$TRAVIS_BUILD_DIR -v /tmp/.X11-unix:/tmp/.X11-unix -w $TRAVIS_BUILD_DIR -e DISPLAY=$DISPLAY -- qrexec-test"
-      script: $DOCKER_RUN python3 -m coverage run -m unittest discover -s qrexec/tests -t . -p '*.py' -v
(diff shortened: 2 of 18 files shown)


marmarek:
I've done some tests. Start situation: qrexec call into dom0 (`/bin/true` as a service) takes ~380ms. Observations:
1. Enabling this policy daemon saves ~100ms
2. `qubes-rpc-multiplexer` takes about ~130ms
3. Disabling policy loading on each call (caching it) saves another ~120ms

The second point should be hopefully solved by #3 (for latency critical services).
The third point could be solved by some clever caching the policy, including detection of changes. Or requiring explicit policy reload after modification, but that wouldn't be nice...

marmarek:
Some issues detected during tests:
- [ ] `qubes-qrexec-policy-daemon` service is not set to autostart (missing `enable` entry in a preset file)
- [ ] socket is created with wrong permissions (0644 instead of 0660, especially group qubes don't have access, making it impossible to connect)
- [ ] logs about allowed connections are duplicated in journalctl - once as `qrexec` and then as `qrexec-policy-daemon`
- [ ] the daemon waits for `qrexec-client` process to terminate before serving next requests; in case of connection to dom0, this means waiting until the whole connection is terminated; this can be also an issue (untested) for calls into DisposableVM, where the daemon needs to cleanup the VM after connection is terminated

DemiMarie:
@marmarek @marmarta is there still a reason to fall back to the old solution?

systemd/qubes-qrexec-policy-daemon.service
```diff
@@ -0,0 +1,7 @@
+[Unit]
+Description=Qubes remote exec policy daemon
+After=qubesd.service
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/qrexec-policy-daemon
```
DemiMarie: It would be nice if this used socket activation.

marmarek:
> @marmarek @marmarta is there still a reason to fall back to the old solution?

Yes, if policy daemon isn't running/can't start for any reason. Having some kind of fallback is especially useful with GUI VM (qrexec as the only interface to dom0). This is also why socket activation may not be the best idea - it would prevent such fallback from working. On the other hand, automatic restart on crash may be worth adding.

marmarek:
I'll merge this PR only after https://github.com/QubesOS/qubes-core-qrexec/pull/3, as conflict resolution will be much easier this way.

PR #11, opened 2019-11-10
Add README

pierwill:
Is line 1 accurate?


marmarek:
Yes, looks fine. But take a look at https://www.qubes-os.org/doc/code-signing/ - for code repositories we require commits to be signed.

PR #12, opened 2019-12-10
WIP: follow up to pr#3 (socket based services)

HW42:
This is a follow-up to @DemiMarie's PR #3.

The fixes are probably best integrated into PR #3. I will add a review comment there.

It's still WIP. The biggest point will be probably rebasing on the new revision of PR #3. To prepare for this I tried not include cleanup changes but only required changes. This hopefully will lead to much fewer merge conflicts.

Another problem: If the service closes the write part of the unix socket this is not detected so it will hang until the caller tries to write something and triggers a `SIGPIPE`. Do we want or need to support half-closed unix sockets? If yes how do you detect `shutdown(s, SHUT_WR)`? The manpage of `poll` suggest that this will show `POLLERR` in the `revents` but in my tests this didn't work (but a write triggered `SIGPIPE` as expected).


marmarek:
> Another problem: If the service closes the write part of the unix socket this is not detected so it will hang until the caller tries to write something and triggers a `SIGPIPE`. Do we want or need to support half-closed unix sockets?

Just a clarification, you meant `shutdown(s, SHUT_RD)` done by the service, right?
I don't think that's an issue. It is much more important (and enough) to detect the other direction closed - service telling it's end of data it want to send - either by doing `close(s)` or `shutdown(s, SHUT_WR)`.

HW42:
> Just a clarification, you meant shutdown(s, SHUT_RD) done by the service, right?

Yes, sorry.

> I don't think that's an issue. It is much more important (and enough) to detect the other direction closed - service telling it's end of data it want to send - either by doing close(s) or shutdown(s, SHUT_WR).

So you propose that for unix socket services we will interpret an EOF when reading from the socket (i.e. the service called SHUT_WR or close) as end of the service, i.e. we also close the other direction and send a fake exit code, correct? If I'm not mistaken for non socket services half open connection are currently fully supported.

marmarek:
> So you propose that for unix socket services we will interpret an EOF when reading from the socket (i.e. the service called SHUT_WR or close) as end of the service,  i.e. we also close the other direction and send a fake exit code, correct?

No, I propose to transfer EOF to the other end (like I believe it is already done). And terminate connection (send fake exit code etc) when both ends sends EOF.

HW42:
> No, I propose to transfer EOF to the other end (like I believe it is already done).

If I didn't messed up my tests this is working.

> And terminate connection (send fake exit code etc) when both ends sends EOF.

The problem is that the other EOF is not send in some cases.

For example if a VM is running `socat tcp-listen:1234,fork 'exec:qrexec-client-vm dst-vm some.service'` socat doesn't close if it gets an EOF from qrexec-client-vm. So even if the unix socket in dst-vm is completely closed it will still wait (because it thinks it could write stuff).

marmarek:
> socat doesn't close if it gets an EOF from qrexec-client-vm

Can you provide more details? I think socat should shutdown its writing end only when it gets EOF on the actual connection (TCP in this case), no? It should be like this:
0. (setup, socat call)
1. In `src-vm`: `nc localhost 1234` 
2. Connection is made to `some.service` in `dst-vm` (where `some.service` can be socket-based, on exec based)
3. `some-service` uses shutdown() to signal EOF
4. `socat` gets EOF, can act on it or not (for example in case of TCP, send FIN packet - socat does that I think(*))
5. the remote end of (TCP) connection gets EOF, can still send some data
6. the remote end of (TCP) connection sends EOF, closing TCP connection completely
7. socat transfers EOF to qrexec-client-vm (either by shutdown(), or close() at this point)
8. qrexec transfers EOF to the other end
9. since both directions are closed at this point, fake exit code is produced and connection is closed

Steps 3-5 and 6-8 can be swapped and it should still work.

And FWIW, we do have automated tests for both cases, and also for VM-dom0/VM-VM: https://github.com/QubesOS/qubes-core-admin/blob/master/qubes/tests/integ/vm_qrexec_gui.py#L167-L270 (and one of them is marked with "expected failure"...)

In case where EOF isn't handled by the client side (for example socat translating to a protocol without means to transfer it), there will probably need to be some in-band EOF-equivalent or timeout or whatnot. But that's outside of qrexec scope.

(*) It looks like socat at this point impose some timeout - if no data in the other direction is sent anymore, it closes the connection after about half a second. But if data is still going, the connection stays half-open and the data is transferred. `nc` seems to not transfer EOF in some cases.

HW42:
Ok, I think we have been thinking about two different scenarios (I was looking at the "the service dies" case). But let's first handle your scenario as outlined above (Maybe this solves the other case good enough).

Having looked into this I'm currently stuck at this point:

> 4. `socat` gets EOF, can act on it or not [...]

This doesn't seem to work as expected, but it doesn't look like a qrexec problem. Simplified test case:

```
cat > close-stdout3.c <<EOF
#include <unistd.h>
#include <stdio.h>

int main() {
    dprintf(2, "writing something\n");
    dprintf(1, "test\n");
    usleep(1000000);
    dprintf(2, "closing stdout\n");
    close(1);
    usleep(5000000);
    dprintf(2, "bye\n");
}
EOF
gcc -Wall close-stdout3.c -o close-stdout3
socat -d -d -d - exec:./close-stdout3
```
(Using tcp-listen instead of stdio as other end doesn't seem to change things)

Output:
```
2019/12/11 14:59:35 socat[16274] I socat by Gerhard Rieger and contributors - see www.dest-unreach.org
2019/12/11 14:59:35 socat[16274] I This product includes software developed by the OpenSSL Project for use in the OpenSSL Toolkit. (http://www.openssl.org/)
2019/12/11 14:59:35 socat[16274] I This product includes software written by Tim Hudson (tjh@cryptsoft.com)
2019/12/11 14:59:35 socat[16274] N reading from and writing to stdio
2019/12/11 14:59:35 socat[16274] N forking off child, using socket for reading and writing
2019/12/11 14:59:35 socat[16274] I socketpair(1, 1, 0, {5,6}) -> 0
2019/12/11 14:59:35 socat[16274] N forked off child process 16275
2019/12/11 14:59:35 socat[16275] I just born: child process 16275
2019/12/11 14:59:35 socat[16275] I close(4)
2019/12/11 14:59:35 socat[16275] I close(3)
2019/12/11 14:59:35 socat[16275] I close(5)
2019/12/11 14:59:35 socat[16275] I dup2(6, 0) -> 0
2019/12/11 14:59:35 socat[16275] I dup2(6, 1) -> 1
2019/12/11 14:59:35 socat[16275] I close(6)
2019/12/11 14:59:35 socat[16275] N execvp'ing "./close-stdout3"
2019/12/11 14:59:35 socat[16274] N forked off child process 16275
2019/12/11 14:59:35 socat[16274] I close(6)
2019/12/11 14:59:35 socat[16274] I resolved and opened all sock addresses
2019/12/11 14:59:35 socat[16274] N starting data transfer loop with FDs [0,1] and [5,5]
writing something
test
2019/12/11 14:59:35 socat[16274] I transferred 5 bytes from 5 to 1
closing stdout
bye
2019/12/11 14:59:41 socat[16274] N childdied(): handling signal 17
2019/12/11 14:59:41 socat[16274] I childdied(signum=17)
2019/12/11 14:59:41 socat[16274] I exec'd process 16275 on socket 1 terminated
2019/12/11 14:59:41 socat[16274] I waitpid(): child 16275 exited with status 0
2019/12/11 14:59:41 socat[16274] I waitpid(-1, {}, WNOHANG): No child processes
2019/12/11 14:59:41 socat[16274] I childdied() finished
2019/12/11 14:59:41 socat[16274] N socket 2 (fd 5) is at EOF
2019/12/11 14:59:41 socat[16274] I terminated child did not leave data for us
2019/12/11 14:59:42 socat[16274] I poll timed out (no data within 0.500000 seconds)
2019/12/11 14:59:42 socat[16274] I shutdown(5, 2)
2019/12/11 14:59:42 socat[16274] N exiting with status 0
```

So it seems socat doesn't get an EOF? But this works:

```
cat > eof-test <<EOF
#!/usr/bin/env python3

from subprocess import Popen, PIPE
import time

p = Popen(["./close-stdout3"], stdout=PIPE, stdin=PIPE)

while True:
    d = p.stdout.read(1)
    print(repr(d))
    if d == b'':
        break
time.sleep(10)
EOF
chmod +x eof-test
./eof-test
```

Output:
```
writing something
b't'
b'e'
b's'
b't'
b'\n'
closing stdout
b''
bye
```

HW42:
PS: If socat is using pipes or a socket to communicate with it's child doesn't matter:
```
$ socat -d -d -d - exec:./close-stdout3,pipes
2019/12/11 15:09:29 socat[16298] I socat by Gerhard Rieger and contributors - see www.dest-unreach.org
2019/12/11 15:09:29 socat[16298] I This product includes software developed by the OpenSSL Project for use in the OpenSSL Toolkit. (http://www.openssl.org/)
2019/12/11 15:09:29 socat[16298] I This product includes software written by Tim Hudson (tjh@cryptsoft.com)
2019/12/11 15:09:29 socat[16298] N reading from and writing to stdio
2019/12/11 15:09:29 socat[16298] I setting option "pipes" to 1
2019/12/11 15:09:29 socat[16298] N forking off child, using pipes for reading and writing
2019/12/11 15:09:29 socat[16298] I pipe({5,6}) -> 0
2019/12/11 15:09:29 socat[16298] I pipe({7,8}) -> 0
2019/12/11 15:09:29 socat[16298] N forked off child process 16299
2019/12/11 15:09:29 socat[16298] N forked off child process 16299
2019/12/11 15:09:29 socat[16298] I resolved and opened all sock addresses
2019/12/11 15:09:29 socat[16298] N starting data transfer loop with FDs [0,1] and [5,8]
2019/12/11 15:09:29 socat[16299] I just born: child process 16299
2019/12/11 15:09:29 socat[16299] I close(4)
2019/12/11 15:09:29 socat[16299] I close(3)
2019/12/11 15:09:29 socat[16299] I close(5)
2019/12/11 15:09:29 socat[16299] I close(8)
2019/12/11 15:09:29 socat[16299] I dup2(6, 1) -> 1
2019/12/11 15:09:29 socat[16299] I close(6)
2019/12/11 15:09:29 socat[16299] I dup2(7, 0) -> 0
2019/12/11 15:09:29 socat[16299] I close(7)
2019/12/11 15:09:29 socat[16299] N execvp'ing "./close-stdout3"
writing something
test
2019/12/11 15:09:29 socat[16298] I transferred 5 bytes from 5 to 1
closing stdout
bye
2019/12/11 15:09:35 socat[16298] N childdied(): handling signal 17
2019/12/11 15:09:35 socat[16298] I childdied(signum=17)
2019/12/11 15:09:35 socat[16298] I exec'd process 16299 on socket 1 terminated
2019/12/11 15:09:35 socat[16298] I waitpid(): child 16299 exited with status 0
2019/12/11 15:09:35 socat[16298] I waitpid(-1, {}, WNOHANG): No child processes
2019/12/11 15:09:35 socat[16298] I childdied() finished
2019/12/11 15:09:35 socat[16298] I poll(): Interrupted system call
2019/12/11 15:09:35 socat[16298] I terminated child did not leave data for us
2019/12/11 15:09:36 socat[16298] I poll timed out (no data within 0.500000 seconds)
2019/12/11 15:09:36 socat[16298] I close(5)
2019/12/11 15:09:36 socat[16298] N exiting with status 0
```

marmarek:
>     dprintf(2, "closing stdout\n");
>     close(1);

If socket is duplicated on multiple FD, only the last close will do a thing. `shutdown(1, SHUT_WR)` should work. This is BTW one of the motivations to use sockets instead of pipes to communicate with child process in qrexec - otherwise it's impossible to half-close the connection in multi-process case (something called from a script).

> Ok, I think we have been thinking about two different scenarios (I was looking at the "the service dies" case).

As I said, exactly the same should work in this case - order of EOFs should matter.
In short: service dies, so qrexec will get EOF (all copies of that socket closed), and will transfer it to the other end. Now the other end can either:
 - try to read something - will get EOF
 - try to write something - will get fake exit code after qrexec trying to write to the service and failing, and also eventually EPIPE (when writing again to now terminated qrexec-client-vm)

There is a corner case when the client doesn't read nor write - just keep the connection open, but I think we can ignore this case (i.e. keep the connection open, or in fact half-closed, even though the service end is dead).
Another corner case is caused by qrexec buffering - client writing to the (dead) service - the first write will succeed, as qrexec don't see that direction closed until failed write. But then again, the client in practice will either want to write more data (which will fail this time), or read a response (and will get EOF).

DemiMarie:
> And FWIW, we do have automated tests for both cases, and also for VM-dom0/VM-VM: https://github.com/QubesOS/qubes-core-admin/blob/master/qubes/tests/integ/vm_qrexec_gui.py#L167-L270 (and one of them is marked with "expected failure"...)

If the problem is what I think it is, my PR should fix it.  I remember fixing at least one problem in the qrexec core.  I don’t remember what it is, though.

DemiMarie:
@HW42 if you can sign your commits I can merge them into my branch.

HW42:
> @HW42 if you can sign your commits I can merge them into my branch.

Fixed.

DemiMarie:
@HW42 I still get that one of the commits does not have a GPG signature.

HW42:
> @HW42 I still get that one of the commits does not have a GPG signature.

The head (and thereby the history) of my branch is commit [66599bc5](https://github.com/HW42/qubes-core-qrexec/commit/66599bc53605463695b551b36dfd4b0d1b3b14ad) which is singed by tag [hw42_66599bc5](https://github.com/HW42/qubes-core-qrexec/releases/tag/hw42_66599bc5)

What's missing?

marmarek:
Merged within #3

PR #17, opened 2020-02-05
Add vchan-socket based tests

pwmarcz:


marmarek:
there is already `qrexec/tests` for various tests (mostly policy parser), can you reuse that place?

pwmarcz:
How about now? I moved them and made `./run-tests` run everything.

(I tried to pull in Pytest already but ran into some issues with `qrexec/tests/cli.py`, and I don't want to make this change any bigger).

daemon/qrexec-daemon.c
```diff
@@ -256,53 +261,60 @@ void init(int xid)
...
-         0660);
-
-    if (logfd < 0) {
-        perror("open");
-        exit(1);
-    }
 
-    dup2(logfd, 1);
-    dup2(logfd, 2);
+    if (getuid() == 0) {
```
marmarek: Currently qrexec-daemon is _not_started as root, so this isn't a good condition here (it is good only in qrexec-agent). Maybe put it under `-D` option too (daemon -> log to file, no daemon -> log to stderr)?

pwmarcz: Yeah, I think it's better than another option. Done.

marmarek: You've pushed to #19 branch (which is fine and also needed), but not here. I'm ok with that, but github won't notice when it's merged.

PR #18, opened 2020-02-05
Separate GTK test classes from object under test

pwmarcz:
The tests for GTK objects used multiple inheritance to combine
TestCase class and the class being tested. This made the code more
confusing, but also caused the GTK objects to be initialized when
loading (not executing) the test, which can cause a segfault if
X is not available.

I rewrote the tests to use composition instead of inheritance,
creating an object (self.window, etc.) in the setUp method.

### DIFF
--- a/qrexec/tests/gtkhelpers.py
+++ b/qrexec/tests/gtkhelpers.py
@@ -18,6 +18,8 @@
 # License along with this library; if not, see <https://www.gnu.org/licenses/>.
 #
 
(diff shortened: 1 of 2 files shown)


pwmarcz:
Build failure looks kind of random, I'll re-trigger:

>fatal: unable to access 'https://github.com/QubesOS/qubes-builder-rpm.git/': Failed to connect to github.com port 443: Connection timed out

PR #19, opened 2020-02-06
daemon: don't listen to clients while reconnecting to agent 

pwmarcz:
Otherwise, connections from client will hang.

See QubesOS/qubes-issues#5347.


marmarek:
Oops, here travis failures weren't random.

PR #22, opened 2020-02-08
Collect test coverage data about C programs too

marmarek:
This improves tests coverage collection.

Note it include also two commits from #3, to fix Makefile ignoring LDLIBS.
Preferably it should be merged after #3 (and rebased on top).


pwmarcz:
Now that #3 is merged, I will try to add some more graceful SIGTERM handling to daemon and agent. Agent's children are long running and can be abandoned, but daemon connections (from client) are short-lived so we can wait for children there.

marmarek:
I've added some already here, quite minimal, do you want to do it better?

pwmarcz:
Oh, I missed that and tried to implement it by myself. One thing that came up: does the SIGTERM handling work when the daemon is trying to reconnect? I added a test (`test_terminate_before_restart` in #24) that hangs on your code.

In the case of socket-based version, it hangs inside `libvchan_client_init` (there is `connect` and `nanosleep` in a loop there). But I think the Xen version might have a similar problem: `libvchan_client_init` is supposed to wait for the server, and we have no guarantee the signal will come in the right moment for something to fail with `EINTR`.

Maybe resume the normal behavior in that part? Something like:

```c
signal(SIGTERM, SIG_DFL);
// In case the signal handler fired just before
if (terminate_requested)
    return -1;

libvchan_client_init(...);

signal(SIGTERM, sigterm_handler);
```

I don't have a better idea, from what I see calling `exit()` is unsafe in a signal handler :(

marmarek:
> does the SIGTERM handling work when the daemon is trying to reconnect?

Not really. And this is also why I modified the test tear down to first terminate the daemon.

On Xen, it is less of an issue, because libvchan_client_init() will eventually give up when the target domain die. But in vchan-socket case, we don't have such thing.

I've considered aborting libvchan_client_init() when the the socket's directory disappears, but I think still it wouldn't work with the current test anyway.

pwmarcz:
Then, what do you think about the above, i.e. default SIGTERM behaviour during reconnect? It sounds better than leaving the process hanging indefinitely.

marmarek:
We won't collect coverage data from that test then... Can we do that, _and_ avoid killing daemon while reconnecting in tests?

pwmarcz:
I think we should keep at least one test that checks this exact thing; i.e. whether the process reacts to `SIGTERM` when reconnecting. There is also another `test_agent_reconnect` which is a normal test for the reconnecting functionality, and it finishes after the agent reconnects, so no coverage from that function would be lost. Would that be OK?

I guess we could also add `__gcov_flush` there (under a necessary `#ifdef`), but that sounds very hacky and probably not worth it...

PR #24, opened 2020-02-17
Socket based qrexec: refactor file search, add tests

pwmarcz:
This is on top of QubesOS/qubes-core-qrexec#3. Main part is in 36ef73421484.

### DIFF
--- a/libqrexec/exec.c
+++ b/libqrexec/exec.c
@@ -29,6 +29,7 @@
 
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <sys/stat.h>
 #include <sys/un.h>
 #include <sys/wait.h>
 #include <unistd.h>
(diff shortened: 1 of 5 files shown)


libqrexec/exec.c
```diff
@@ -272,107 +303,89 @@ int execute_qubes_rpc_command(char *cmdline, int *pid, int *stdin_fd,
...
-    size_t const service_length = delimiter ?
-        (size_t)(delimiter - command->service_descriptor) : command->service_descriptor_length;
-
-    if (!service_length) {
+    if (!command->service_descriptor_length) {
         fputs("Service path empty\n", stderr);
         return -1;
-    } else if (service_length > NAME_MAX) {
-        fprintf(stderr, "Service path too long to execute: %zu\n", service_length);
+    } else if (command->service_descriptor_length > NAME_MAX) {
```
marmarek: This will break services where service name is ok, but its argument is long.

pwmarcz: Right. I reverted that part.

pwmarcz:
`test_run_dom0_command_and_connect_vm` seems to fail randomly with the `qrexec-client` disconnecting too soon, both here and on master. I made the test a bit weaker and added a `TODO`.

PR #25, opened 2020-02-20
Qrexec policy daemon: solve merge conflicts, update

pwmarcz:
For pull request #6.


pwmarcz:
The failing test seems to be the same as in #24, I cherry-picked the workaround here.

PR #28, opened 2020-02-24
Various fixes for socket handling corner cases

marmarek:
Result of debugging session together with @pwmarcz

May fix QubesOS/qubes-issues#5597


marmarek:
BTW, while debugging this, we've found this python issue: https://bugs.python.org/issue39744

pwmarcz:
Tests, and most of the fixes, applied in #33.

PR #29, opened 2020-02-25
Switch policy-agent to qrexec

pwmarcz:
Requires QubesOS/qubes-core-admin#323.

PR #30, opened 2020-02-28
Add notifications (allowed, denied, failed)

pwmarcz:
See QubesOS/qubes-issues#3904.

Based on #29.


qrexec/policy/parser.py
```diff
@@ -523,7 +525,8 @@ def from_ask_resolution(cls, ask_resolution, *, target):
             if dispvm:
                 self.cleanup_dispvm(target)
         if process.returncode != 0:
-            raise AccessDenied('qrexec-client failed')
+            # Return "request refused" to the remote end.
+            raise AccessDenied('qrexec-client failed', notify=False)
```
marmarek: Shouldn't this be `notify=True` (as with other denied cases)?

pwmarcz: This is a confusing corner case, because this happens after we showed an "Allowed" notification, so the user might see both. Do you think that's better than failing silently?

pwmarcz:
One more thing - should we maybe wrap the notify call in an exception handler and proceed anyway? Right now, if qrexec-policy-agent doesn't work, calls get blocked. Then again, only the calls that the user explicitly asked to be notified about.

marmarek:
If just notification is broken, I think it should log an error but continue anyway. If ask itself fails, then obviously deny the call.

marmarek:
Also, please document policy.* calls API. It should definitely be listed at https://www.qubes-os.org/doc/vm-interface/#qubes-rpc - but I think it would make sense to just link from there to a new section on https://www.qubes-os.org/doc/qrexec-internals/

qrexec/policy/parser.py
pwmarcz: I thought about this and I actually think we should not notify here.

1. `qrexec-client` can also return a non-zero exit code in case the call went through, but the program on remote end exited with non-zero status (though not always: in some cases `qrexec-client` returns early). So, while we could add a special notification ("qubes.Service failed"), it would be pretty confusing, because it wouldn't be shown in every case.

2. This is pure speculation but source program might expect a non-zero exit code as normal part of operation, so it's not even a failure (imagine "qvm-run grep ..."). In any case, I think it's better to leave handling the failure to the source.

3. The situation this piece of code is actually supposed to handle is when something goes wrong with setting up the connection. I think ensuring that the source gets `MSG_CONNECTION_REFUSED` is right, but treating this situation as a "acces denied" isn't.

So, for now I would use another exception here (`raise ConnectionFailed`?) to make it clear that this is not a policy refusal. And leave a comment explaining that this is a workaround and we don't currently distinguish cases 1/2 from case 3.

marmarek: This is indeed confusing corner case. The issue with just `MSG_CONNECTION_REFUSED` and no notification is that the source side may expect the notification was already shown for `MSG_CONNECTION_REFUSED`, so doing own error notification would be wrong in most cases.

Maybe better solution would be checking if target's qrexec-daemon is alive at the final policy evaluation step (when we know the actual target already), before answering "allow" at all? It may not handle all the failure modes, but the most common will and will also avoid confusing qrexec-client exit code cases.
In the future (as discussed somewhere already), we should move VM<->VM connection negotiation to the qrexec-daemon itself, so qrexec-client's only responsibility will be connecting local process (or its own stdin/stdout) with vchan and the confusing exit code case should be gone.

pwmarcz: Checking if the daemon is alive would mean trying to connect to it, right? Because the daemon process might be even alive but in a "reconnecting" state where it doesn't accept connections.

With how the code is laid out currently, the current connection happens too late. We could connect early and pass the FD to qrexec-client. Alternatively, maybe add a "ping the daemon" step to the policy-daemon, that just opens a connection to see if it would be accepted? (To be honest both solutions seem kind of ugly).

marmarek: What about reserving yet another exit code (125?) in qrexec-client for "connection failed" and raise exception only in that case?

pwmarcz: I like it. I guess it might collide with service's exit code, but it sounds unlikely.

I would still display a special notification ("resolution": "connection_failed", not "resolution": "deny"; and another message for the user).

marmarek: Yes, different notification sounds good.

pwmarcz: Okay, no, I don't like it anymore :( There are too many `exit()`s in qrexec-client and I really don't want to add another "feature" to it right now.

pwmarcz: Oh, but the other way around is easier - I can add an option to exit with 0 even if child/remote exited with non-zero.

pwmarcz:
image
image

qrexec/server.py
```diff
@@ -0,0 +1,118 @@
...
+    qrexec_opts = ['-d', remote_domain]
+    cmd = 'DEFAULT:QUBESRPC {} {}'.format(service, source_domain)
+    command = [QREXEC_CLIENT] + qrexec_opts + [cmd]
+    process = await asyncio.create_subprocess_exec(
+        *command,
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE)
+    stdin, _stderr = await process.communicate(
+        json.dumps(params).encode('ascii'))
+    return stdin.decode('ascii')
```
marmarek: There is already `qrexec.client_dom0` and `qrexec.client_vm` (and a code in __init__.py to choose one based on where it is called). They are different (not async def, no json encoding, no direct socket support for local calls), so not exactly duplication, but IMO worth keeping them in the same place, and document differences, to avoid confusion. Also, distinguishing local from remote based on target being dom0 works only in dom0...

pwmarcz: I extracted the generic part to `qrexec.client_dom0` and `qrexec.client_vm`, and documented the limitations of current module.

Documentation/qrexec-policy-agent.md
```diff
@@ -0,0 +1,34 @@
...
+
+The policy agent works as a socket-based service. It handles the following Qubes RPC calls:
+
+### `policy.Ask`
+
+Display a prompt about whether to allow an action.
+
+The request is a JSON dictionary with the following keys:
+- `source` - source domain (`"work"`)
+- `service_name` - name of service without argument (`"qubes.Filecopy"`)
```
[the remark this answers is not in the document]
pwmarcz: Okay, somehow I thought it's without argument because of [the discussion about notifications](https://github.com/QubesOS/qubes-issues/issues/3904#issuecomment-592516200). I will change that.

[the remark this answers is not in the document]
pwmarcz: Actually, that seems to be a regression in 4.1, introduced in 2018 with the new version of the tool. The old qrexec-policy [used the service_name argument directly](https://github.com/QubesOS/qubes-core-admin/blob/release4.0/qubespolicy/cli.py#L44), the new tool strips the argument during parsing.

PR #33, opened 2020-03-05
Common main loop for agent and client

pwmarcz:
What remains is the main loop; for now it's pretty different in both versions.

### DIFF
--- a/daemon/qrexec-client.c
+++ b/daemon/qrexec-client.c
@@ -50,7 +50,8 @@ static pid_t local_pid = 0;
 /* flag if this is "remote" end of service call. In this case swap STDIN/STDOUT
  * msg types and send exit code at the end */
 static int is_service = 0;
-static int child_exited = 0;
+
+static volatile sig_atomic_t sigchld = 0;
 
 static const char *socket_dir = QREXEC_DAEMON_SOCKET_DIR;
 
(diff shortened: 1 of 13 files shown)


pwmarcz:
@marmarek A few of your commits still remain, especially the deadlock fix, c9abe6c07a0852dd82fbc7af8a4ce8c52d91f8ec (but I don't understand it well enough to be sure of it).

pwmarcz:
Main loop replaced, deadlock fix applied :)

libqrexec/libqrexec-utils.h
```diff
@@ -98,4 +101,96 @@ static inline size_t max_data_chunk_size(int protocol_version) {
...
+      is_service true (this is a service):
+        - send local data as MSG_DATA_STDOUT
+        - send exit code
+        - wait just for local end
+        - return local exit code
+
+      is_service false (this is a client):
+        - send local data as MSG_DATA_STDIN
+        - don't send exit code
+          (TODO: should it be this way?)
```
marmarek: Yes.

libqrexec/libqrexec-utils.h
```diff
@@ -98,4 +101,96 @@ static inline size_t max_data_chunk_size(int protocol_version) {
...
+        - if there's no process, wait for remote end
+        - return local or remote exit code, whichever finished first
+          (TODO: should it be this way?)
```
marmarek: It should not depend on their exit order. As for which exit code return - the reasoning for returning local one (if local process present) is that local process may have better chance to interpret remote data and handle various expected failures gracefully (but it have no way to receive remote process exit code directly - can only use in-band data). On the other hand, this sometimes returning remote exit code and sometimes not is pretty confusing, so it may worth unifying it (and always return remote exit code).
I don't think this change would break anything, as local process is used rarely. But if we'd go with the change, we need to verify this.

daemon/qrexec-client.c
```diff
@@ -322,315 +275,34 @@ static void send_service_connect(int s, char *conn_ident,
...
+    req.stderr_fd = -1;
+    req.local_pid = local_pid;
+    req.is_service = is_service;
+    req.replace_chars_stdout = replace_chars_stdout;
+    req.replace_chars_stderr = replace_chars_stderr;
+    req.data_protocol_version = data_protocol_version;
+    req.sigchld = &sigchld;
+    req.sigusr1 = NULL;
+
+    fprintf(stderr, "calling process_io\n");
```
marmarek: Debug leftover.

marmarek:
Based on recent test results, I'm going to merge this one in its current shape. There are some improvements that could be made, but it will be easier in a separate PR.

PR #41, opened 2020-03-25
Support wait-for-session in qrexec-client

pwmarcz:


libqrexec/exec.c
```diff
@@ -262,78 +252,205 @@ static int find_qrexec_service_file(
...
+                    plus - start);
+                goto err;
+            }
+            cmd->service_name = strndup(start, plus - start);
+            if (!cmd->service_name) {
+                PERROR("strndup");
+                goto err;
+            }
+        } else {
+            cmd->service_name = strdup(cmd->service_descriptor);
```
marmarek:
```suggestion
            cmd->service_name = strndup(cmd->service_descriptor, end - start);
```
?

pwmarcz: I'll do `strndup(start, end - start)`, same as for `cmd->service_descriptor`.

PR #42, opened 2020-03-28
qrexec-client: handle failed service exec

pwmarcz:
Otherwise, qrexec-client just hangs when it fails to spawn a
process, or connect to a socket service.

This corresponds to the fix for qrexec-agent in commit
95fa4bb2d29fcfc91da3ada95300258450022610.

### DIFF
--- a/daemon/qrexec-client.c
+++ b/daemon/qrexec-client.c
@@ -252,18 +252,47 @@ out:
     destroy_qrexec_parsed_command(cmd);
 }
 
-static void prepare_local_fds(char *cmdline, struct buffer *stdin_buffer)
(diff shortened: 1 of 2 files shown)


marmarek:
Do you think it fixes the issue in https://openqa.qubes-os.org/tests/7251 ?
I don't see failed service calls there (at least not that many), but a lot of qrexec-policy-exec processes - that many that dom0 runs out of memory.
It may be that qrexec-policy-exec hangs after service is executed - before the OOM, there was a lot of (successful) calls to `qubes.UpdatesProxy` (from one VM to another).

pwmarcz:
AFAICT, the impact for this one is minimal - only socket services in dom0 that are not running at the moment.

PR #56, opened 2020-10-11
Fix qrexec-agent crashing with "pselect: Bad file descriptor"

marmarek:
See indiviual commit messages for details
Fixes QubesOS/qubes-issues#6120


fepitre:
I've relaunched CI with mirror updated here too.

PR #57, opened 2020-10-11
policy: adjust call target for allow action too

marmarek:
Similarly to 'ask' default_target parameter, expand actual 'target'
parameter in case of allow action (explicit or implicit via ask). This
makes the '@adminvm' keyword isolated to the policy internal logic, but
when communicating externally, it gets translated to the actual VM name
(which currently always is 'dom0', but may be translated into "local
 AdminVM" in the future).

QubesOS/qubes-issues#6112

### DIFF
--- a/qrexec/policy/parser.py
+++ b/qrexec/policy/parser.py
@@ -520,7 +520,7 @@ class AllowResolution(AbstractResolution):
 
         target = self.target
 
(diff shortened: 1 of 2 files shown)


fepitre:
I've relaunched CI with mirror updated. No timeout due to long update.

PR #66, opened 2021-01-20
Allow to build w/o pam

easydozen:
QubesOS/qubes-issues#5802

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ install-dom0: all-dom0
 
 
 all-vm:
-	+$(MAKE) all CFLAGS=-DHAVE_PAM LDLIBS=-lpam -C agent
+	+$(MAKE) all -C agent
 .PHONY: all-vm
 
 install-vm: all-vm
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ install-dom0: all-dom0
 
 
 all-vm:
-	+$(MAKE) all -C agent
+	+$(MAKE) all CFLAGS=-DHAVE_PAM LDLIBS=-lpam -C agent
 .PHONY: all-vm
 
 install-vm: all-vm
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -67,7 +67,7 @@ install-dom0: all-dom0
 
 
 all-vm:
-	+$(MAKE) all -C agent
+	+$(MAKE) all CFLAGS=-DHAVE_PAM LDLIBS=-lpam -C agent
```
marmarek, line 70: You override CFLAGS/LDLIBS here, which breaks reproducible builds (drops flags set by dpkg-buildpackage, rpmbuild etc).
But also, setting LDLIBS in this place looks wrong (see the next comment).

agent/Makefile
```diff
@@ -10,7 +10,7 @@ override QUBES_CFLAGS := -I. -I../libqrexec -g -O2 -Wall -Wextra -Werror \
    -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fPIC -std=gnu11 -D_POSIX_C_SOURCE=200809L \
 	-D_GNU_SOURCE $(CFLAGS)
 override LDFLAGS += -pie -Wl,-z,relro,-z,now -L../libqrexec
-override LDLIBS=-lqrexec-utils $(shell pkg-config --libs $(VCHAN_PKG)) -lpam
```
marmarek, line 13: `-lpam` should be added in this Makefile, not elsewhere. Maybe a better way is to do it conditionally on `-DHAVE_PAM` in `$(CFLAGS)`? Or even separate make variable (`HAVE_PAM`?) to set both `-DHAVE_PAM` in `CFLAGS` and `-lpam` in `LDFLAGS`.

qrexec/tools/qrexec_policy_exec.py
```diff
@@ -117,8 +117,11 @@ async def execute(self, caller_ident):
 
 class NotifyAllowedResolution(parser.AllowResolution):
     async def execute(self, caller_ident):
-        guivm = \
+        try:
+            guivm = \
             self.request.system_info['domains'][self.request.source]['guivm']
+        except Exception:
```
marmarek, line 123: Pylint complains about broad exception here. Should be KeyError I guess.

marmarek (changes requested):

marmarek (approved):

PR #67, opened 2021-03-03
build: RPM spec package debuginfo match the actually created package

kalkin:


marmarek:
Can you explain why exactly this change is needed? The automatically generated debuginfo package seems correct.

kalkin:
You are right the buildinfo file is correct, because I think `rpmbuild` throws magic at it to figure out where actual debuginfo is if you don't provide it. `rpm` does not so if you do `rpm -q -D "dist .fc33" --qf '%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}.src.rpm\n' --specfile` you get wrong results. 

The actual background is that I am working on anather approach to skipping building packages already build when running `make qubes`. For this I need the expected output of querying the spec file should match the buildinfo file.

marmarek:
The automatic debuginfo package generation is specifically to not do it manually, which is prone to errors, easy to desynchronize with the actual package content etc. If some part of the machinery doesn't handle it correctly - that part needs fixing.
That said, we do have some solution for that already - the build saves ".list" files with the built package list. And we have already [a code that checks if it's fresh and use if if so](https://github.com/QubesOS/qubes-builder-rpm/blob/master/Makefile.rpmbuilder#L94-L101). So, perhaps the solution here is to remove the fallback in case of no list file present (or outdated one) and treat that situation as package not built? In fact, I did [exactly this in builder-archlinux](https://github.com/QubesOS/qubes-builder-archlinux/blob/master/Makefile.archlinux#L63-L71). Generating package list based only on a spec file, without .list file is a legacy feature that shouldn't be needed anymore.

kalkin:
@marmarek Thank you for your elaborate answer. I will have a look at builder-archlinux, because currently I'm just sidestepping calling `$(MAKE) -f Makefile.generic PACKAGE=$$package` and wrapp it in to my own makefile. This obviosly works for me but if you are interested in an upstream solution this is not enough.

I come to same conclusion about using `*.list` files.  I noticed that I can't fix all the specfiles.  I'm currently rewriting it :grin: 

About this PR: Dunno :shrug:  Take it or leave it. If the above approach works like it supposed to, I don't care.

Thank you for your feedback
kalkin

marmarek:
What you also may be interested to look at is `make check-release-status` and the machinery behind it. It sounds like a building block for the thing you are trying to do. In fact, currently it is kind of slow exactly because of the fallback for the missing `*.list` file, it should be much faster if we remove that fallback.

As for this PR, I'm closing it for now.

kalkin:
Without this pr the goal `update-repo-do` fails because it searches for wrong debugfile

marmarek:
Is that a case where you didn't rebuilt a package after changing something? Otherwise it should take correct package from a `.list` file.

kalkin:
@marmarek While we are on the topic of taking the correct packages from .list `linux-template-builder` building fails because it can't find some python packages which are actually called `python3-...`. It looks like [this](https://github.com/QubesOS/qubes-builder-rpm/blob/master/Makefile.rpmbuilder#L208) is the issue.

marmarek:
Can you paste specific error message? I can't really tell what issue you have in mind from this description...

kalkin:
```
ln: failed to access 'pkgs/vm-fc33/python-objgraph-3.4.1-2.fc33.noarch.rpm': No such file or directory
make[1]: *** [/home/user/Projects/qubes-builder/qubes-src/builder-rpm/Makefile.rpmbuilder:163: update-repo-do] Error 1
make: *** [Makefile:342: template-local-fc33] Error 1
make linux-template-builder --no-silent VERBOSE=1  21.24s user 12.79s system 144% cpu 23.550 total
```

marmarek:
Do you have python-objgraph built? This looks like yet another case resolving binary packages based purely on spec file is flawed. But using `.list` file should work (as long as you have built the package).

kalkin:
Ahh I found the issue. I don't have the *.list file because i hacked my setup to use `*.buildinfo` files, because this way I can figure out what to download.

PR #69, opened 2021-04-19
Add a policy.Eval qrexec service

DemiMarie:
The current implementation is naive, but a faster one is certainly
possible.

### DIFF
--- a/qrexec/tests/qrexec_policy_daemon.py
+++ b/qrexec/tests/qrexec_policy_daemon.py
@@ -22,7 +22,6 @@ import asyncio
 from contextlib import suppress
 
 import pytest
-import asynctest
 from unittest.mock import Mock
 import functools
 
@@ -35,7 +34,7 @@ from ..tools import qrexec_policy_daemon
 class TestPolicyDaemon:
     @pytest.fixture
     def mock_request(self, monkeypatch):
-        mock_request = asynctest.CoroutineMock()
+        mock_request = unittest.mock.AsyncMock()
         monkeypatch.setattr('qrexec.tools.qrexec_policy_daemon.handle_request',
                             mock_request)
         return mock_request
@@ -57,7 +56,7 @@ class TestPolicyDaemon:
     async def qrexec_server(self, tmp_path, request):
         log = unittest.mock.Mock()
 
-        qrexec_server = await asyncio.start_unix_server(
+        server = await asyncio.start_unix_server(
             functools.partial(qrexec_policy_daemon.handle_qrexec_connection,
                               log, Mock()),
             path=str(tmp_path / "socket.qrexec"))
--- a/qrexec/tests/qrexec_policy_daemon.py
+++ b/qrexec/tests/qrexec_policy_daemon.py
@@ -53,15 +53,32 @@ class TestPolicyDaemon:
 
         server.close()
 
-    async def send_data(self, server, path, data):
+    @pytest.fixture
+    async def qrexec_server(self, tmp_path, request):
+        log = unittest.mock.Mock()
+
+        qrexec_server = await asyncio.start_unix_server(
+            functools.partial(qrexec_policy_daemon.handle_qrexec_connection,
+                              log, Mock()),
+            path=str(tmp_path / "socket.qrexec"))
+
+        yield server
+
+        server.close()
+
+    async def send_data(self, server, path, data, qrexec=False):
         reader, writer = await asyncio.open_unix_connection(
-            str(path / "socket.d"))
+            str(path / "socket.qrexec" if qrexec else "socket.d"))
         writer.write(data)
+        if qrexec:
+            writer.close()
+
         await writer.drain()
 
         await reader.read()
 
-        writer.close()
+        if not qrexec:
+            writer.close()
 
         server.close()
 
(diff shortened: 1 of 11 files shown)


qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +99,78 @@ async def handle_client_connection(log, policy_cache,
...
+                log.warning('Empty string is not a valid service')
+
+            ### SANITIZE BEGIN
+            untrusted_data = untrusted_data[service_descriptor_len + 1:]
+
+            if len(untrusted_data) > 63:
+                log.warning('Request data too long: %d', len(untrusted_data))
+                return
+            split = untrusted_data.find(b'\0')
+            if not (1 <= split <= 31):
```
marmarek, line 138: This could use a comment what it really does.

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +99,78 @@ async def handle_client_connection(log, policy_cache,
...
+                log.warning('Invalid data from qube')
+                return
+            untrusted_source = untrusted_data[:split].decode('ascii', 'strict')
+            untrusted_target = untrusted_data[split + 1:].decode('ascii', 'strict')
```
marmarek, line 142: A much cleaner version would be `untrusted_source, untrusted_target = untrusted_data.split(b'\0', 1)` and then verify results

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +99,78 @@ async def handle_client_connection(log, policy_cache,
...
+    args = {}
+
+    try:
+        untrusted_data = await reader.read(65536)
+        if len(untrusted_data) > 65535:
+            log.error('Request length too long: %d', len(data))
+            return
+        try:
+            # Qrexec guarantees that this will be present
+            service_descriptor_len = untrusted_data.index(b'\0')
```
marmarek, line 114: `service_descriptor, untrusted_data = untrusted_data.split(b'\0', 1)` ?

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +99,78 @@ async def handle_client_connection(log, policy_cache,
...
+            log.warning('Invalid data from qube')
+            return
+
+        result = await handle_request(
+                source=source,
+                intended_target=intended_target,
+                service_and_arg=service_queried.decode('ascii', 'strict'),
+                domain_id = -1,
+                process_ident = -1
+                assume_yes_for_ask=True,
```
marmarek, line 163: For the very case of gui-daemon and clipboard, this will be enough. But for generic case, I'd like to distinguish also ask from allow, and get more details: https://github.com/QubesOS/qubes-issues/issues/6264
Maybe leave it as is for now, but name the service `policy.EvalSimple` and have full `policy.Eval` later?

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +99,78 @@ async def handle_client_connection(log, policy_cache,
...
+        except (ValueError, UnicodeError):
+            log.warning('Invalid data from qube')
+            return
+
+        result = await handle_request(
+                source=source,
+                intended_target=intended_target,
+                service_and_arg=service_queried.decode('ascii', 'strict'),
+                domain_id = -1,
+                process_ident = -1
```
marmarek, line 162: I'm not sure if those are used with `just_evaluate=True`, but they should be strings.

marmarek (changes requested):

qrexec/tools/qrexec_policy_daemon.py
DemiMarie: Good idea.  If we really cared about performance, a faster approach would be to split at `b' '` first, then at `b'\0'`.  But the complexity probably is not worth it.

qrexec/tools/qrexec_policy_daemon.py
DemiMarie: This checks that the length of a qube name is valid, but a comment would still be useful.

DemiMarie (commented):
Thanks for the review @marmarek!

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -107,19 +201,40 @@ async def start_serving(args=None):
     log = logging.getLogger('policy')
     log.setLevel(logging.INFO)
 
+    for i in (args.eval_socket_path, args.clipboard_socket_path,
```
marmarek, line 204: `args.clipboard_socket_path` doesn't exist. `args.gui_socket_path`?

DemiMarie: :woman_facepalming: Will fix

DemiMarie:
> Tests:
> 
> ```
> May 23 00:58:38 dom0 qrexec-policy-daemon[1252]: b'policy.EvalSimple': invalid data from qube
> ```

The GUI daemon was actually fine.  The problem was in the following Python code:

https://github.com/QubesOS/qubes-core-qrexec/blob/d2ed0ecf738512cfc929a37644bffb3e2ee607be/qrexec/tools/qrexec_policy_daemon.py#L117-L124

This assumes that `await StreamReader.read(65536)` will not return until either all 65536 bytes have been read or EOF is reached.  However, this is not actually the case: it will return when *any* bytes have been read from the socket.  In practice, we usually wind up receiving two buffers: one containing the qrexec metadata, and the other containing the actual data from the remote qube.  Therefore, the bug caused us to only read the metadata buffer, and then error out because we thought the remote qube didn’t send any data.  Now fixed.

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +106,114 @@ async def handle_client_connection(log, policy_cache,
...
+            # This is an error because qrexec should forbid this.
+            log.error('%s invoked with incorrect name %s by %s',
+                      service_name, invoked_service, remote_domain)
+            return
+
+        if not service_queried:
+            log.warning('%s: %s used the empty string as a service name',
+                        service_name, remote_domain)
+            return
+        if len(untrusted_data) > 63:
```
marmarek, line 159: In the specific case for gui-domain (which uses only specific vm names), it is true indeed. But not in generic case - qrexec target may be also a `@dispvm:vm-name` string for example.
Anyway, you already applied sensible size limit earlier (and you'll check that again below...). Since this code doesn't use static-sized variables, and also the policy will validate the name against actual list of existing VMs, there is no need to be this strict here. You can drop this check.

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +106,114 @@ async def handle_client_connection(log, policy_cache,
...
+            log.warning('%s: %s sent %d bytes, but the limit is 63',
+                        service_name, remote_domain, len(untrusted_data))
+            return
+        try:
+            untrusted_source, untrusted_target = untrusted_data.split(b'\0', 1)
+
+            # Check that qube name lengths are reasonable
+            # pylint: disable=superfluous-parens
+            if not (1 <= len(untrusted_target) <= 31):
```
marmarek, line 168: See above, this is not necessarily true in generic case. The current version of the qrexec protocol allows up to 63 chars in the target name.

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +106,114 @@ async def handle_client_connection(log, policy_cache,
...
+            sanitize_domain_name(untrusted_source, True)
+            sanitize_domain_name(untrusted_target, True)
+            ### SANITIZE END
+            source, intended_target = untrusted_source, untrusted_target
+        except (ValueError, UnicodeError) as e:
+            log.warning('%s: invalid data from qube %s', service_name, remote_domain)
+            return
+
+        if check_gui:
+            system_info = get_system_info()['domains']
```
marmarek, line 190: This makes the info loaded twice for a single request. I'm not sure what's the best solution, but some is to add `system_info` keyword argument (defaulting to `None`) to the `handle_request` function and make it load the info only if not already provided.

qrexec/tests/qrexec_policy_daemon.py
```diff
@@ -181,3 +218,208 @@ async def test_wrong_arg(self, mock_request, async_server, tmp_path):
...
+                service_and_arg='d', process_ident='0', log=log,
+                assume_yes_for_ask=True, just_evaluate=True,
+                policy_cache=unittest.mock.ANY)
+            mock_request.reset_mock()
+
+    @pytest.mark.asyncio
+    async def test_not_guivm(
+            self, mock_request, qrexec_server, tmp_path, mock_system):
+        data = b'policy.EvalGUI+d b keyword @adminvm\0c\0a'
+        assert len(qrexec_server) == 2
```
marmarek, line 242: Test for testing tests? ;) This would make adding tests for other sockets in the future annoying. Similar in the tests below.

qrexec/tests/qrexec_policy_daemon.py
```diff
@@ -181,3 +218,208 @@ async def test_wrong_arg(self, mock_request, async_server, tmp_path):
...
         mock_request.assert_not_called()
+
+    @pytest.mark.asyncio
+    async def test_simple_qrexec_request(
+            self, mock_request, qrexec_server, tmp_path, mock_system):
+
+        for i in server_types:
+            data = b'policy.Eval%s+d c keyword @adminvm\0a\0b' % i
+
+            assert await self.send_data(qrexec_server[i], tmp_path, data, i) == b'result=deny\n'
```
marmarek, line 229: Add also some test with "allow" result (by adjusting `mock_request`). Can be this one in fact, as you test very similar thing with deny in `test_not_guivm` test for `policy.EvalSimple` case.

marmarek (changes requested):
Besides the comments below, see pylint and pytest output.

qrexec/tools/qrexec_policy_daemon.py
```diff
@@ -99,6 +106,112 @@ async def handle_client_connection(log, policy_cache,
...
+            sanitize_domain_name(untrusted_target, True)
+            ### SANITIZE END
+            source, intended_target = untrusted_source, untrusted_target
+        except (ValueError, UnicodeError):
+            log.warning('%s: invalid data from qube %s', service_name, remote_domain)
+            return
+
+        system_info = get_system_info()
+        if check_gui:
+            system_info = system_info['domains']
```
marmarek, line 187: This overrides `system_info` var, that you then pass into `handle_request`.

marmarek (changes requested):
Besides one comment left from previous review, there is a new one.

marmarek:
```
qrexec/tools/qrexec_policy_daemon.py:123:0: R0914: Too many local variables (21/20) (too-many-locals)
```
:facepalm:

marmarek:
PipelineRetry

marmarek (approved):

PR #72, opened 2021-05-19
Link with -lgcov when building with coverage

DemiMarie:
Otherwise we get linker errors.


marmarek:
Apparently `__gcov_flush` got removed in GCC 11: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93623#c12
I guess we should use `__gcov_dump` instead? Not sure about `__gcov_reset`, but maybe too?

marmarek (approved):

PR #74, opened 2021-05-21
Create a DispVMTemplate instance when needed

DemiMarie:
Otherwise we assert when trying to launch the DispVM.  The assertion is
still useful to catch bugs elsewhere in the system.

Fixes QubesOS/qubes-issues#6629.

### DIFF
--- a/qrexec/policy/parser.py
+++ b/qrexec/policy/parser.py
@@ -504,6 +504,8 @@ class AllowResolution(AbstractResolution):
     @classmethod
     def from_ask_resolution(cls, ask_resolution, *, target):
         '''This happens after user manually approved the call'''
+        if target.startswith('@dispvm:'):
+            target = DispVMTemplate(target)
         return cls(
             ask_resolution.rule,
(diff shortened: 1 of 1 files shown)


marmarek:
1. Your assert fired in tests
2. use `isinstance` instead of `type(...) is` - see pylint.

DemiMarie:
> 1. Your assert fired in tests
> 2. use `isinstance` instead of `type(...) is` - see pylint.

I removed the assertion altogether.  It was intended to ensure that the prompting code was actually returning a `str` (and not a subclass), but clearly at least the test suite breaks that assumption.

marmarek:
It isn't just tests, see this:
```
class AssumeYesForAskResolution(parser.AskResolution):
    async def execute(self, caller_ident):
        return await self.handle_user_response(
            True, self.request.target).execute(caller_ident)
```

`self.request.target` is a `IntendedTarget` object.

marmarek (approved):

PR #81, opened 2021-11-07
Fix man page format

marmarek:
Fixes QubesOS/qubes-issues#7041


marmarek:
It's better but still doesn't render properly.

PR #83, opened 2021-12-10
Avoid leaking file descriptors

DemiMarie:
This mostly avoids spurious SELinux denials, but it is good practice anyway.

Should fix the file descriptor leakage in #5214.

### DIFF
--- a/agent/qrexec-agent.c
+++ b/agent/qrexec-agent.c
@@ -29,9 +29,11 @@
 #include <unistd.h>
 #include <stddef.h>
 #include <errno.h>
+#include <dirent.h>
 #include <err.h>
 #include <sys/wait.h>
 #include <fcntl.h>
+#include <sys/ioctl.h>
 #include <string.h>
 #include <getopt.h>
 #include <pwd.h>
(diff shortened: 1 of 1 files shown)


agent/qrexec-agent.c
```diff
@@ -119,6 +121,43 @@ static struct pam_conv conv = {
...
+        long fd_num;
+        if (!entry) {
+            if (errno)
+                err(1, "readdir(/proc/self/fd)");
+            break;
+        }
+        errno = 0;
+        fd_num = strtol(entry->d_name, &endptr, 10);
+        if (errno || !endptr || *endptr || fd_num < 0 || fd_num > INT_MAX)
+            errx(1, "bad number in /proc/self/fd");
```
marmarek, line 149: This always trips, I'm pretty sure it's about `.` or `..`.

Anyway, setting all FDs to `O_CLOEXEC` sounds fragile (there may be cases when inheriting open FD would be desirable, even if currently it isn't used), and also could hide other issues. I'd prefer to do it properly:
 - inspect what FDs are currently passed on to children, and for each consciously decide whether `O_CLOEXEC` is desirable (or maybe it shouldn't be left open at all?)
 - (optionally) add a test that checks for FD leaks - we do have proper unit tests in this repo, so it shouldn't be too hard. see `qrexec/tests/socket/agent.py` for example

marmarek (changes requested):

PR #86, opened 2021-12-13
Check the return value of snprintf

DemiMarie:
It will fail if the output buffer is too small or if the implementation tries and fails to allocate heap memory.


DemiMarie:
Tested and works on my machine.

marmarek (approved):

PR #96, opened 2022-09-14
Prefix data support

DemiMarie:
This allows passing a prefix data argument to qrexec-client-vm.  Prefix
data is sent before the data on stdin, which is useful for e.g.
requesting armored or binary OpenPGP signatures.


DemiMarie:
PipelineRetryFailed

marmarek (approved):

PR #97, opened 2022-09-21
Only spawn services via exec

DemiMarie:
Non-executable services are rejected by qrexec already.


DemiMarie:
PipelineRetryFailed

lib/qubes-rpc-multiplexer
```diff
@@ -51,14 +51,5 @@ for DIR in $QREXEC_SERVICE_PATH; do
 done
 IFS=$ifs
 
-if [ -x "$CFG_FILE" ] ; then
-    # shellcheck disable=SC2086
-	exec "$CFG_FILE" ${QREXEC_SERVICE_ARGUMENT}
-	echo "$0: failed to execute handler for" "$1" >&2
```
marmarek, line 57: You removed this error message, intentional?

marmarek (changes requested):

lib/qubes-rpc-multiplexer
DemiMarie: Yes; most shells never return from a failed `exec`.

marmarek: I don't think so:
```
$ test -x /tmp/error.sh; echo $?
0
$ exec /tmp/error.sh; echo $?
bash: /tmp/error.sh: /bin/no-such-file: bad interpreter: No such file or directory
bash: /tmp/error.sh: Success
126
$ cat /tmp/error.sh
#!/bin/no-such-file
```

lib/qubes-rpc-multiplexer
```diff
@@ -51,14 +51,7 @@ for DIR in $QREXEC_SERVICE_PATH; do
...
-else
-    # shellcheck disable=SC2086
-	exec /bin/sh -- "$CFG_FILE" ${QREXEC_SERVICE_ARGUMENT}
-	echo "$0: failed to execute handler for" "$1" >&2
-	exit 1
-fi
+# shellcheck disable=SC2086
+exec "$CFG_FILE" ${QREXEC_SERVICE_ARGUMENT}
+echo "$0: failed to execute handler for $1" >&2
+exit 126
```
marmarek, line 57: 126 is used by qrexec-client-vm for "request refused". Please try to avoid conflicting with it. You can simply preserve the old value (1).

marmarek (changes requested):

marmarek (approved):

PR #98, opened 2022-11-23
Reject bad messages

DemiMarie:
In R4.1 these are allowed and ignored.  Reject them in R4.2.


DemiMarie:
PipelineRetryFailed

DemiMarie:
PipelineRetryFailed

marmarek:
Even the current version is too strict to preserve compatibility (https://github.com/QubesOS/qubes-issues/issues/7917). Yes, technically it is a bug on the QWT side, but the change breaks something that was working before. This PR is going even further, without much practical benefit, so I'm going to reject it. We can reconsider it for Qubes 5.x (closest time for backward-incompatible changes, no timeline for it yet).

PR #99, opened 2022-11-27
Relax validation of MSG_DATA_STDIN/MSG_DATA_STDOUT

marmarek:
Since they are handled the same way, do not ignore remote side sending
the wrong one, just warn about it (once). QWT have it mixed up, so
ignoring such messages breaks communication with QWT.

Fixes QubesOS/qubes-issues#7917


DemiMarie:
Perhaps also `Fixes: 6daa16d6b079070d69754ac9cfa0ed3cd77a3d85 ("Prevent log injection into qrexec-agent")`, if we wish to record the commit that introduced the problem?

DemiMarie (approved):

PR #100, opened 2022-12-08
Enable `-D_FORTIFY_SOURCE=3`

DemiMarie:
This improves `_FORTIFY_SOURCE` coverage on compilers that support it.


marmarek:
Same reason as in https://github.com/QubesOS/qubes-core-vchan-xen/pull/30

PR #104, opened 2023-03-27
Add types to Python code

DemiMarie:
No functional change intended.  Some code has not been typed yet.

### DIFF
--- a/qrexec/policy/parser.py
+++ b/qrexec/policy/parser.py
@@ -22,6 +22,7 @@
 
 """Qrexec policy parser and evaluator"""
 
+from __future__ import annotations
 import abc
 import collections
 import collections.abc
@@ -34,13 +35,26 @@ import string
 import asyncio
 
 from typing import (
+    Any,
+    Collection,
+    FrozenSet,
+    Generator,
     Iterable,
     List,
+    NoReturn,
+    Sequence,
+    Set,
     TextIO,
     Tuple,
+    TYPE_CHECKING,
+    Type,
     Dict,
     Optional,
+    Union,
 )
+
+if TYPE_CHECKING:
+    from _typeshed import Self
 
 from .. import QREXEC_CLIENT, POLICYPATH, RPCNAME_ALLOWED_CHARSET, POLICYSUFFIX
 from .. import exc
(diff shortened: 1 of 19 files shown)


qrexec/policy/utils.py
```diff
@@ -19,13 +19,23 @@
...
-import pyinotify
+import pyinotify # type: ignore
 from qrexec import POLICYPATH, POLICYPATH_OLD
+from typing import Optional, List, Dict
 from . import parser
 
 
 class PolicyCache:
-    def __init__(self, path=POLICYPATH, use_legacy=True, lazy_load=False):
+    path: str
```
marmarek, line 29: POLICYPATH is `pathlib.Path`, not `str`

qrexec/policy/utils.py
```diff
@@ -44,11 +54,11 @@ def __init__(self, path=POLICYPATH, use_legacy=True, lazy_load=False):
         self.watches = []
         self.notifier = None
 
-    def initialize_watcher(self):
-        self.watch_manager = pyinotify.WatchManager()
+    def initialize_watcher(self) -> None:
+        self.watch_manager: pyinotify.WatchManager = pyinotify.WatchManager()
```
marmarek, line 58: (here, but also as a general remark) I don't think annotation on assignment is necessary when it's exactly the type of what's being assigned (and it isn't somehow obfuscated). Especially when the variable you assign to is already annotated. In fact, it might be even harmful, as it might hide mismatch between such annotations.

qrexec/tools/qrexec_policy_agent.py
```diff
@@ -321,10 +322,23 @@ def can_perform_action(self):
 
 class RPCConfirmationWindow:
     # pylint: disable=too-few-public-methods,too-many-instance-attributes
-    _source_file = pkg_resources.resource_filename(
+    __slots__ = ('_gtk_builder',
```
marmarek, line 325: Please avoid using `__slots__`. Generally it isn't pythonic approach, and it exists only when memory usage is critical for large data amounts.

qrexec/policy/parser.py
```diff
@@ -550,7 +550,7 @@ class AbstractResolution(metaclass=abc.ABCMeta):
     """Object representing positive policy evaluation result -
     either ask or allow action"""
 
-    def __init__(self, rule, request, *, user):
+    def __init__(self, rule, request: Request, *, user: str):
```
marmarek, line 553: `Request` isn't declared at this point. But you can use [forward reference](https://peps.python.org/pep-0484/#forward-references) (`'Request'`), or `from __future__ import annotations`

PR #105, opened 2023-04-02
Miscellaneous cleanups

DemiMarie:
See individual commit messages for details.
Fixes https://github.com/QubesOS/qubes-issues/issues/6628


DemiMarie:
PipelineRetry

marmarek:
Buth pylint and pytest are not happy yet.

And my attempt to run openQA on this revealed another issue, with upgrade path: VMs still running old qrexec-daemon while both qrexec-policy-daemon and qrexec-policy-exec are new ones will fail to run any call. This is only one time issue, but it will prevent clean in-place upgrade (you can't run `qubes-dom0-update` after upgrading qrexec-daemon). I think the easiest solution would be to:
1. Make sure old qrexec-daemon communicating with new policy daemon do not understand the answer (maybe even let the new qrexec-daemon include some extra field in the query, that new policy daemon expects and otherwise refuses to process the request). This should trigger fallback to qrexec-policy-exec.
2. Add a flag to qrexec-policy-exec that it's called by the new qrexec-daemon. If the flag is missing, process using the old approach (call qrexec-client from there).

marmarek:
You probably seen that, but both pylint and pytest says the things is very broken.

run-tests
```diff
@@ -7,14 +7,14 @@ fi
...
+    make -C daemon BACKEND_VMM=socket clean all
 else
     echo "libvchan-socket not available, skipping socket tests"
     export SKIP_SOCKET_TESTS=1
 fi
 
 set -x
 
-python3 -m coverage run -m pytest -o python_files=*.py -v qrexec/tests "$@"
+python3 -m coverage run -m pytest -o python_files=*.py -v "$@"
```
marmarek, line 20: If you change this, add the argument you removed to the call in `.gitlab-ci.yml`, otherwise it collects unintended files.

marmarek:
The integration test (with still old qrexec-daemon running) say:
```
  File "/usr/bin/qrexec-policy-exec", line 5, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_exec.py", line 304, in main
    result = get_result(args)
             ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_exec.py", line 282, in get_result
    if intended_target[0] == '@':
       ^^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'intended_target' where it is not associated with a value
```
I'm not sure why pylint hasn't complained, but also this part is not covered by any unit test.

run-tests
```diff
@@ -1,20 +1,30 @@
 #!/bin/bash -e
 
 case $0 in (/*) cd "${0%/*}/";; (*/*) cd "./${0%/*}";; esac
+if [[ "$#" -eq 0 ]]; then
+    set -- qrexec/tests
+else
+    for i; do
+        case $i in
+        (qrexec/tests|qrexec/tests/*) :;;
+        (*) printf 'Unsupported path %q\n' "$i" >&2; exit 1;;
```
marmarek, line 10: Please don't. This breaks using any options for pytest via the script.

marmarek (changes requested):
(partial review)

marmarek:
pylint found legitimate issue

run-tests
marmarek: I guess I wasn't clear enough. Do not add arguments validation to this script. This still breaks several cases (like options with arguments). And I forbid you to re-implement pytest arguments parser here.

daemon/qrexec-daemon.c
```diff
@@ -752,87 +782,200 @@ static void sanitize_name(char * untrusted_s_signed, char *extra_allowed_chars)
...
+                            request_id->ident,
+                            NULL);
+                    PERROR("execl");
+                } else {
+                    PERROR("snprintf");
+                }
+                _exit(126);
+            default:
+                if (close(fds[1]))
+                    abort();
```
marmarek, line 860: If you move the fallback to a separate function and wrap response parsing in a loop (executed at most twice), I think you can rather easily have it also in case of invalid response too (which is IMO useful thing to have).

daemon/qrexec-daemon.c
```diff
@@ -853,37 +996,81 @@ static void handle_execute_service(
...
+        if (strncmp("@dispvm:", target, sizeof("@dispvm:") - 1) == 0) {
+            disposable = true;
+            buf = qubesd_call(target + 8, "admin.vm.CreateDisposable", "", &resp_len);
+            if (!buf) // error already printed by qubesd_call
+                _exit(126);
+            if (resp_len >= 2 && memcmp(buf, "0", 2) == 0) {
+                /* we exec later so memory leaks do not matter */
+                target = buf + 2;
+            } else {
+                LOG(ERROR, "qubesd could not create disposable VM");
```
marmarek, line 1035: Log returned error message from qubesd too?

marmarek (changes requested):

marmarek:
pylint found some minor issues, but also one unit test failed (unsure why)

qrexec/policy/parser.py
```diff
@@ -962,17 +962,19 @@ def evaluate(self, request):
                     self.rule.filepath, self.rule.lineno
                 )
             )
-        if target == "@dispvm":
-            target = target.get_dispvm_template(  # pylint: disable=no-member
+        if isinstance(target, DispVM):
+            target_ = target.get_dispvm_template(  # pylint: disable=no-member
```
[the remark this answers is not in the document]
marmarek: I see, apparently it can't :/

qrexec/policy/parser.py
```diff
@@ -465,48 +502,55 @@ class DispVM(Target, Redirect, IntendedTarget):
...
 
     @staticmethod
-    def get_dispvm_template(source, *, system_info):
+    def get_dispvm_template(source: str, *, system_info: SystemInfo) -> Optional[DispVMTemplate]:
         """Given source, get appropriate template for DispVM. Maybe None."""
-        if (
-            source not in system_info["domains"]
-            or system_info["domains"][source]["default_dispvm"] is None
-        ):
+        template = system_info[source].get("default_dispvm", None)
```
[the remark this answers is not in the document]
DemiMarie: Done.

qrexec/tools/qrexec_policy_exec.py
```diff
@@ -367,9 +368,9 @@ async def handle_request(
             )
         )
         resolution = policy.evaluate(request)
-        v = await resolution.execute()
+        result = await resolution.execute()
         assert isinstance(v, str)
```
marmarek, line 372: you haven't updated variable name here...

marmarek (changes requested):
Something went wrong and now the main change is squashed into `Add types to Python code` commit, with the original commit message gone.

Just in case that it was:
```
Avoid having qrexec-policy-daemon handle connections
     
This makes qrexec-policy-daemon stateless and ensures that restarting it
does not e.g. cause data sent over qrexec to be corrupted.  There should
also be a performance improvement.

Most of the decrease in lines of code is because lots of the Python
tests simply do not make sense any more, as they test code that no
longer exists.  Unit-testing C code is significantly more difficult and
so the current plan is to rely on integration tests.

Known limitations:
- Notifications for VM start failures have not been implemented.
- If qrexec-policy-daemon returns an invalid response (as opposed to
  refusing the connection) there will be no fallback to
  qrexec-policy-exec.

Old versions of qrexec-daemon will still need qrexec-policy-exec to run
qrexec-client.  Add a unit test for this.
```

qrexec/utils.py
```diff
@@ -117,12 +131,12 @@ def get_system_info():
     """
 
     system_info = qubesd_call("dom0", "internal.GetSystemInfo")
-    return json.loads(system_info.decode("utf-8"))
+    return cast(SystemInfo, json.loads(system_info.decode("utf-8"))["domains"])
```
marmarek, line 134: I had a bad feeling about this change and now I know why. The policy modules are used in other places too, for example [here](https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/ext/admin.py#L76). So, this API change (expected layout of `system_info`) broke them. While it's theoretically okay to change the other places too (not too many of them), there is also another concern: this `domains` was put there for a reason - to leave a place if ever policy would need some more info (like global properties). I don't think there is much performance impact with this extra level, but also I see it will make the annotations a bit more complex. So, if I haven't missed any more important reason to eliminate it, I'd prefer to leave it.

marmarek (changes requested):

marmarek:
I added ticket reference to the PR description, please add also to relevant commit(s) messages.

DemiMarie:
> I added ticket reference to the PR description, please add also to relevant commit(s) messages.

Done

marmarek:
> * TC_00_Qrexec_debian-11: [test_090_qrexec_service_socket_dom0](https://openqa.qubes-os.org/tests/71544#step/TC_00_Qrexec_debian-11/16) (failure)
>       `AssertionError: b'test.Socket test-inst-vm1 keyword adminvm\x000123...`

Service name is supposed to always have `+` included, even if the caller did not provided a service argument.

DemiMarie:
> > * TC_00_Qrexec_debian-11: [test_090_qrexec_service_socket_dom0](https://openqa.qubes-os.org/tests/71544#step/TC_00_Qrexec_debian-11/16) (failure)
> >   `AssertionError: b'test.Socket test-inst-vm1 keyword adminvm\x000123...`
> 
> Service name is supposed to always have `+` included, even if the caller did not provided a service argument.

What is the reason for this?  In the old code it appears to have been accidental rather than intentional.  I fixed this in the most recent force-push by using `strchr()` to check if the name includes `+`, but it seems to me that “no argument” vs “empty argument” could be useful information to pass.

marmarek:
> it seems to me that “no argument” vs “empty argument” could be useful information to pass.

Policy-wise they are the same (intentionally, to reduce confusion and avoid subtle errors - both in policy and services themselves), so the service call should also get the same regardless of the form used by the caller.

marmarek:
> Service name is supposed to always have + included, even if the caller did not provided a service argument.

This is still an issue, I think the condition is reversed.

libqrexec/ioall.c
```diff
@@ -90,6 +98,66 @@ int read_all(int fd, void *buf, int size)
...
+    }
+    char *buf = malloc(buf_size);
+    if (buf == NULL) {
+        LOG(ERROR, "malloc() for %zu bytes failed!", buf_size);
+        abort();
+    }
+    *len = 0;
+    for (;;) {
+        size_t to_read = buf_size - 1 - offset;
+        ssize_t res = read(fd, buf + offset, to_read > INT_MAX ? INT_MAX : to_read);
```
marmarek, line 121: I'd prefer to cap `max_bytes` on INT_MAX (this function is ever called with max_bytes at most 64k), than trying to handle such corner case.

marmarek (changes requested):

libqrexec/ioall.c
DemiMarie: Done, along with adding lots of comments to this function.

libqrexec/ioall.c
```diff
@@ -90,6 +98,103 @@ int read_all(int fd, void *buf, int size)
...
+            *len = offset;
+            break;
+        }
+        if (bytes_read_this_time > to_read)
+            abort(); // kernel bug or some sort of corruption
+        if (bytes_read_this_time == to_read) {
+            /* Buffer full.  See if a new buffer can be allocated. */
+            if (buf_size >= max_bytes) {
+                /* Nope, limit reached. */
+                LOG(ERROR, "Too many bytes read (limit %zu)", max_bytes);
```
marmarek, line 158: Shouldn't you print `max_bytes-1`? Reading exactly `max_bytes` is already rejected (and rightly so, as you no longer subtract 1 from the `buf_size` for the terminating NUL character).

marmarek:
A bunch of manual tests found one more issue: now original target (when calling into dom0) is really what the caller provided, without normalization done by the policy engine. This means several things:
 - non-existing names are no longer replaced with `@default`
 - (a subset of the above) empty name is not replaced with `@default`, breaking such call (qubes-rpc-multiplexer get wrong number of parameters then)
 - dom0 is no longer replaced with `@adminvm` (so, the service would need to handle both forms, if it cares about that)

To avoid inconsistencies, I think policy daemon needs to return that parameter to the qrexec-daemon too.

DemiMarie:
> A bunch of manual tests found one more issue: now original target (when calling into dom0) is really what the caller provided, without normalization done by the policy engine. This means several things:
> 
> * non-existing names are no longer replaced with `@default`
> * (a subset of the above) empty name is not replaced with `@default`, breaking such call (qubes-rpc-multiplexer get wrong number of parameters then)
> * dom0 is no longer replaced with `@adminvm` (so, the service would need to handle both forms, if it cares about that)
> 
> To avoid inconsistencies, I think policy daemon needs to return that parameter to the qrexec-daemon too.

The latter two cases are trivial enough I will handle them in C, but you are correct about the first.

daemon/qrexec-daemon.c
```diff
@@ -746,93 +775,269 @@ static void sanitize_name(char * untrusted_s_signed, char *extra_allowed_chars)
...
     int policy_pending_slot;
     pid_t pid;
-    char remote_domain_id_str[10];
-    sigset_t sigmask;
+
+    switch (target_domain[0]) {
+        case 'A' ... 'Z':
+        case 'a' ... 'z':
+            if (strcmp(target_domain, "dom0") == 0)
+                target_domain = "@adminvm";
```
marmarek, line 1030:
```
Apr 29 01:24:22 dom0 qrexec-policy-daemon[1425]: qrexec: test.AnyvmDeny: test-inst-vm -> @adminvm: denied: no matching rule found
Apr 29 01:24:22 dom0 kernel: qrexec-daemon[17571]: segfault at 0 ip 0000751c2e41362a sp 00007ffc0e519198 error 4 in libc.so.6[751c2e2c8000+156000] likely on CPU 0 (core 0, socket 0)
Apr 29 01:24:22 dom0 kernel: Code: 8b 04 82 29 c8 c3 66 2e 0f 1f 84 00 00 00 00 00 f3 0f 1e fa 89 f1 89 f8 48 83 e1 3f 48 83 e0 3f 83 f9 30 77 3b 83 f8 30 77 36 <f3> 0f 6f 0f f3 0f 6f 16 66 0f ef c0 66 0f 74 c1 66 0f 74 ca 66 0f
```
I'm not sure where the issue is (and whether is it in this line at all), but since you need the requested target transformed by the policy daemon, you can very well drop doing this in qrexec-daemon to simplify the C part and have it done on one place, not two.

marmarek (changes requested):

daemon/qrexec-daemon.c
```diff
@@ -853,37 +1058,105 @@ static void handle_execute_service(
...
+    size_t resp_len;
+    /*
+     * If there was no service argument, pretend that an empty argument was
+     * provided by appending "+" to the service name.
+     */
+    const char *const trailer = strchr(service_name, '+') ? "" : "+";
+
+    /* Check if the target is dom0, which requires special handling. */
+    bool target_is_dom0 = strcmp(target, "@adminvm") == 0 ||
+                          strcmp(target, "dom0") == 0;
```
marmarek, line 1081: The crash is because `target_domain` is NULL at this point. Which is expected for denied call. It should have been aborted earlier, `parse_policy_response()` lost the final
```
if (result != 0)
   _exit(126);
```
In fact, it seems to completely ignore actual allow/deny response, as long as it has correct syntax...

marmarek (changes requested):

marmarek (approved):

PR #109, opened 2023-04-26
Allow specifying a username in service config

DemiMarie:
This also dramatically improves the configuration parser.  Configuration files now use a strict subset of TOML rather than an ad-hoc format with no validation.

Fixes: QubesOS/qubes-issues#6354
Fixes: QubesOS/qubes-issues#8153

### DIFF
--- a/libqrexec/toml.c
+++ b/libqrexec/toml.c
@@ -1,0 +1,276 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <assert.h>
+#include <errno.h>
+#include <stddef.h>
+#include <limits.h>
+
+#include "libqrexec-utils.h"
+
+// A trivial parser for a subset of TOML
+static bool qubes_isspace(unsigned char c) {
+    return c == ' ' || c == '\t';
+}
+
+enum toml_type {
+    TOML_TYPE_INVALID, // error
+    TOML_TYPE_BOOL,
+    TOML_TYPE_INTEGER,
+    TOML_TYPE_STRING,
+};
+
+union toml_data {
+    char *string; // allocated with malloc()
+    bool boolean;
+    unsigned long long integer;
+};
+
+static enum toml_type parse_toml_value(
+    const char *file,
+    size_t line,
+    unsigned char *value_to_parse,
+    union toml_data *value) {
+    enum toml_type ty = TOML_TYPE_INVALID;
+    char *end = NULL;
+
+    switch (value_to_parse[0]) {
+    case '0':
+        ty = TOML_TYPE_INTEGER;
+        value->integer = 0;
+        end = (char *)(value_to_parse + 1);
+        break;
+    case '1' ... '9':
+        ty = TOML_TYPE_INTEGER;
+        errno = 0;
+        value->integer = strtoull((char *)value_to_parse, &end, 10);
+        if (errno) {
+            PERROR("%s:%zu: strtoull()", file, line);
+            return TOML_TYPE_INVALID;
+        }
+        break;
+    case 't':
+        ty = TOML_TYPE_BOOL;
+        if (strncmp((char *)value_to_parse, "true", 4) == 0) {
+            value->boolean = true;
+            end = (char *)(value_to_parse + 4);
+            break;
+        } else {
+            LOG(ERROR, "%s:%zu: Unexpected unquoted string", file, line);
+            return TOML_TYPE_INVALID;
+        }
+    case 'f':
+        ty = TOML_TYPE_BOOL;
+        if (strncmp((char *)value_to_parse, "false", 5) == 0) {
+            value->boolean = false;
+            end = (char *)(value_to_parse + 5);
+            break;
+        } else {
+            LOG(ERROR, "%s:%zu: Unexpected unquoted string", file, line);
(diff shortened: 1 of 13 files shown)


marmarek:
Fuzzer makefile needs an update.

marmarek:
PipelineRetry

marmarek:
> Fuzzer makefile needs an update.

This is still the case: fuzzer fails to link:
```
/usr/bin/ld: libqrexec-exec.o: in function `load_service_config':
/builds/QubesOS/qubes-core-qrexec/fuzz/../libqrexec/exec.c:273: undefined reference to `qubes_toml_config_parse'
```

marmarek:
I may be missing something, but does it actually change the user used for the call? IIUC it only uses the `user` setting to send it to the `qubes.WaitForSession` service, which BTW doesn't even read that value anymore.
The actual user is set in `parse_qubes_rpc_command()` and used in `execute_qubes_rpc_command() -> execute_parsed_qubes_rpc_command()`

marmarek:
> I may be missing something, but does it actually change the user used for the call? IIUC it only uses the `user` setting to send it to the `qubes.WaitForSession` service, which BTW doesn't even read that value anymore.
> The actual user is set in `parse_qubes_rpc_command()` and used in `execute_qubes_rpc_command() -> execute_parsed_qubes_rpc_command()`

This still applies.

DemiMarie:
> > I may be missing something, but does it actually change the user used for the call? IIUC it only uses the `user` setting to send it to the `qubes.WaitForSession` service, which BTW doesn't even read that value anymore.
> > The actual user is set in `parse_qubes_rpc_command()` and used in `execute_qubes_rpc_command() -> execute_parsed_qubes_rpc_command()`
> 
> This still applies.

`handle_server_exec_request_init()` calls `load_service_config()` and (if it picks a user) overwrites the user set by `parse_qubes_rpc_command()`.

marmarek:
Every single test failed on setup with `qrexec-client.c:497:wait_for_vchan_client_with_timeout: vchan connection timeout`. I don't see anything obvious in VM's journal, but it might be in VM's .xsession-errors which wasn't saved.

DemiMarie:
> Every single test failed on setup with `qrexec-client.c:497:wait_for_vchan_client_with_timeout: vchan connection timeout`. I don't see anything obvious in VM's journal, but it might be in VM's .xsession-errors which wasn't saved.

I’ll try installing in a StandaloneVM and see how things break.

marmarek:
Any update on this?

qrexec/tests/socket/agent.py
```diff
@@ -210,6 +210,7 @@ def test_trigger_service(self):
         dom0.handshake()
 
         client = self.connect_client()
+        #input()
```
marmarek, line 213: leftover debugging

marmarek:
There is a merge conflict now :(

libqrexec/toml.c
```diff
@@ -0,0 +1,275 @@
...
+            LOG(ERROR, "%s:%zu: Unterminated quoted string", file, line);
+            return TOML_TYPE_INVALID;
+        } else {
+            size_t to_alloc = (size_t)(end - (char *)value_to_parse);
+            value->string = malloc(to_alloc);
+            if (value->string == NULL) {
+                LOG(ERROR, "%s:%zu: Out of memory copying value", file, line);
+                return TOML_TYPE_INVALID;
+            }
+            memcpy(value->string, value_to_parse, to_alloc - 1);
```
marmarek, line 92:
```suggestion
            memcpy(value->string, value_to_parse + 1, to_alloc - 1);
```
(copy the string without opening quote, and don't truncate its last char)

libqrexec/toml.c
```diff
@@ -0,0 +1,275 @@
...
+                break;
+            case TOML_TYPE_STRING:
+                *user = value.string;
+                continue;
+            default:
+                abort();
+            }
+            LOG(ERROR, "%s:%zu: %s not valid for user name or user ID", config_full_path, lineno, bad_type);
+        } else {
+            LOG(ERROR, "%s:%zu: Unsupported key %s", config_full_path, lineno, current_line);
```
marmarek, line 265: TBH, I'm not sure if unknown settings should fail the call, it will make future extensions harder since often the same qrexec service can be used across many qubes versions. Take this PR as an example - a service that wants to support both old and new version could use the `user` feature, but in its absence fallback to re-exec itself via sudo.

libqrexec/toml.c
```diff
@@ -0,0 +1,275 @@
...
+                    *wait_for_session = (int)value.integer;
+                    continue;
+                }
+                LOG(ERROR, "Integer value %llu used when a boolean was expected", value.integer);
+            } else if (ty == TOML_TYPE_STRING) {
+                LOG(ERROR, "String value '%s' not valid for 'wait-for-session'", value.string);
+                free(value.string);
+                value.string = NULL;
+            }
+        } else if (strcmp(current_line, "user") == 0) {
```
marmarek, line 241: Maybe better call it `force-user`, since it will override `qvm-run --user` argument?

marmarek (changes requested):

marmarek (approved):

PR #114, opened 2023-05-18
Banish old-style function declarations

DemiMarie:
They are deprecated in all versions of C and are fundamentally unsafe. Also ensure that all objects with external linkage are declared before being defined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 MAKEFLAGS=-r
 CC ?= gcc
+CFLAGS += -Werror=strict-prototypes -Werror=old-style-definition -Werror=missing-declarations -Werror=missing-prototypes
 PYTHON ?= python3
-export PYTHON GCC MAKEFLAGS
+export PYTHON CC MAKEFLAGS CFLAGS
 
 SYSLIBDIR ?= /lib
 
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 MAKEFLAGS=-r
-CC ?= gcc
+CC ?= gcc -Werror=strict-prototypes -Werror=old-style-definition -Werror=missing-declarations -Werror=missing-prototypes
```
marmarek, line 2: Flags should not be part of `CC` variable, `CFLAGS` is for this.

marmarek (changes requested):

marmarek:
Compiler found few more cases to fix: https://gitlab.com/QubesOS/qubes-core-qrexec/-/jobs/4338722024#L3056

daemon/qrexec-daemon.c
```diff
@@ -81,54 +81,54 @@ struct _policy_pending {
...
-volatile int child_exited;
-volatile int terminate_requested;
+static volatile int children_count;
+static volatile int child_exited;
+static volatile int terminate_requested;
 
-libvchan_t *vchan;
-int protocol_version;
+static libvchan_t *vchan;
+static int protocol_version;
```
marmarek, line 122: This (and few others) goes beyond the scope you described. And it breaks linking the fuzzer.

marmarek (changes requested):

marmarek (approved):

PR #115, opened 2023-05-23
Ensure that QSB-089 can never happen again

DemiMarie:
It turns out that gcc is able to validate that the QSB-089 fixes worked and optimize accordingly.  __attribute__((error)) can be used to generate a compile-time error if a call to a function remains after optimization, and one can use this to validate that the fix to QSB-089 worked.  Ensures that if the code is vulnerable to QSB-089, it will not compile.

### DIFF
--- a/daemon/qrexec-daemon.c
+++ b/daemon/qrexec-daemon.c
@@ -1196,6 +1196,7 @@ static void sanitize_message_from_agent(struct msg_header *untrusted_header)
                     "although it uses protocol %d", protocol_version);
                 exit(1);
             }
+            /* add 1 for NUL terminator */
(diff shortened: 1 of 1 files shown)


marmarek:
I don't like this at all. It basically duplicate `sanitize_message_from_agent` (so, a risk to get this code desynchronized, if somebody ever update one instance but not the other), but also, it very much depends on behavior of specific compiler in specific version (including its optimizations), which _much_ more likely will backfire than actually detect an issue.

A proper place for such check would be an unit test for this issue. But do not read it as a request to write one, it's (now) well covered by fuzz testing (and I did verified that the current updated fuzzing actually does find issue like this).

PR #117, opened 2023-05-26
Use a pipe instead of a socket for stderr

DemiMarie:
qubes-rpc-multiplexer gives the child a FIFO as stderr, so this should not be a compatibility problem.  Also use XDG_RUNTIME_DIR (if present) in preference to /tmp.


marmarek:
This makes `test_service_close_stdout_stderr_early` failing reliably.

DemiMarie:
That is interesting.  I will diagnose this eventually.

PR #119, opened 2023-06-02
Reject obviously bogus service names

DemiMarie:
A service name that contains a NUL byte is clearly bogus.  Reject the service call rather than truncating the service name.  Also reject service names that start with `+`, as this corresponds to the name of the service (rather than its argument) being the empty string.  I don’t think the new policy format can express this.

### DIFF
--- a/daemon/qrexec-daemon.c
+++ b/daemon/qrexec-daemon.c
@@ -1340,21 +1340,26 @@ void handle_message_from_agent(void)
                 LOG(ERROR, "Service name contains NUL byte at offset %zu", nul_offset);
                 goto fail;
             }
-            if (service_name_len < 1) {
+            switch (untrusted_service_name[0]) {
+            case '\0':
                 LOG(ERROR, "Empty service name not allowed");
                 goto fail;
-            }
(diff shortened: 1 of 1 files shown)


daemon/qrexec-daemon.c
```diff
@@ -1323,26 +1323,49 @@ void handle_message_from_agent(void)
...
             }
+            service_name_len -= 1;
 
             /* sanitize start */
             ENSURE_NULL_TERMINATED(untrusted_params3.target_domain);
-            untrusted_service_name[service_name_len-1] = 0;
             sanitize_name(untrusted_params3.target_domain, "@:");
-            sanitize_name(untrusted_service_name, "+");
             if (!validate_request_id(&untrusted_params3.request_id, "MSG_TRIGGER_SERVICE3")) {
-                free(untrusted_service_name);
```
marmarek, line 1333: Removing this will leak `untrusted_service_name` - `free` part handles only `service_name` which is `NULL` at this point. Similar for other cases you add in this patch below.

marmarek (changes requested):

marmarek (approved):

PR #121, opened 2023-08-03
Ensure that building qrexec-agent builds libqrexec

DemiMarie:
Otherwise the build fails with an extremely obscure linker error that gives absolutely no clue what the actual problem is.  This change also enables building packages manually via rpmbuild, which is much faster than using qubes-builder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ install-dom0: all-dom0
 .PHONY: install-dom0
 
 
-all-vm:
+all-vm: all-base
 	+$(MAKE) all -C agent
 all-vm-selinux:
 	+$(MAKE) -f /usr/share/selinux/devel/Makefile -C selinux qubes-core-qrexec.pp


marmarek:
Note that `all-base`, `all-dom0` and `all-vm` are built as _separate packages_. Building base when just `all-vm` is requested is wasteful. And in fact breaks the build (as requires more build deps that aren't declared in -vm package).
Maybe you didn't have installed (or updated, if you made some ABI change) `qubes-core-qrexec-devel` or `-libs` package? They are properly declared as build dependencies.

DemiMarie:
If I build with `rpmbuild` (as opposed to Qubes Builder, which is far, _far_ slower), `all-vm` gets built against the already installed libqrexec, which fails with a confusing error.  Is there a way to make Qubes Builder as fast (about 11 seconds) as raw `rpmbuild`?

marmarek:
> `all-vm` gets built against the already installed libqrexec

Yes, exactly. If you run `rpmbuild` directly, it's supposed to use installed build deps. Making it build against something else is a bug. If you want to use plain `rpmbuild`, you need to install proper build dependencies yourself in that VM (which probably is not a good idea in some cases - that's why `mock` and similar exist).

> Is there a way to make Qubes Builder as fast (about 11 seconds) as raw `rpmbuild`?

No, I don't think so. But also, if you want a quick build done not in a proper environment, you can very well use `make` directly (and maybe `make install`, after copying the source dir to the target VM?).

PR #123, opened 2023-08-29
Add a tool to convert legacy policy files to new format

marmarta:
fixes QubesOS/qubes-issues#8000

### DIFF
--- a/qrexec/tools/qrexec_legacy_convert.py
+++ b/qrexec/tools/qrexec_legacy_convert.py
@@ -40,7 +40,6 @@ SERVICE_TO_FILE = {
     'qubes.UpdatesProxy': '50-config-updates',
     'qubes.OpenInVM': '50-config-openinvm',
     'qubes.OpenURL': '50-config-openurl',
-    'qubes.Gpg': '50-config-splitgpg',
     'qubes.InputKeyboard': '50-config-input',
     'qubes.InputMouse': '50-config-input',
     'qubes.InputTablet': '50-config-input',
@@ -53,6 +52,11 @@ DISCLAIMER = """
 # Policy rules in this file were automatically converted from old
 # policy format. Old files can be found in /etc/qubes-rpc/policy; they have
 # the suffix .rpmsave.
+
+# All rules pertaining to qubes.Gpg were moved to this file. If you want to
+# move them to the relevant GUI config file, see 50-config-splitgpg. Caution:
+# the GUI tool for qubes.Gpg policy supports only policy rules that have a
+# single vm as a target.
 """
 
 TOOL_DISCLAIMER = """
@@ -94,7 +98,8 @@ class RuleWrapper:
             if 'allow' not in str(self.rule.action):
                 return False
             return self.rule.target.type != 'keyword' or \
-                str(self.rule.target) == '@adminvm'
+                str(self.rule.target) == '@adminvm' or \
+                str(self.rule.target) == '@default'
 
         if self.service in ('qubes.OpenInVM', 'qubes.OpenURL'):
             if self.rule.target != '@dispvm':
@@ -108,15 +113,8 @@ class RuleWrapper:
                 target = '@dispvm'
             return '@dispvm' in target
 
-        if self.service == 'qubes.Gpg':
-            if self.rule.source == '@anyvm':
-                if self.rule.target.type == 'keyword':
-                    # not supported main rule
-                    return False
-            return True
-
         if self.service == 'policy.RegisterArgument':
-            return self.rule.argument == '+u2f.Register'
+            return self.rule.argument == '+u2f.Authenticate'
 
         if self.service in ['u2f.Authenticate', 'u2f.Register']:
             if self.rule.target.type == 'keyword' and \
@@ -264,7 +262,7 @@ def main(args=None):
                 current_text = file.read_text()
                 if current_text.startswith(TOOL_DISCLAIMER):
                     current_text = current_text[len(TOOL_DISCLAIMER):]
-                text = text.rstrip('\n') + current_text
+                text = text.rstrip('\n') + '\n' + current_text.lstrip('\n')
 
         print("Writing " + str(file) + '...')
         file.write_text(text)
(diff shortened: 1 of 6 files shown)


qrexec/tools/qrexec_legacy_convert.py
```diff
@@ -0,0 +1,240 @@
...
+            if rule.rule.target != '@adminvm':
+                rules_to_save[CONFIG_FILE].append(rule)
+            elif rule.rule.action.user:
+                rules_to_save[CONFIG_FILE].append(rule)
+            else:
+                rules_to_save[filename].append(rule)
+            if missing:
+                rules_to_save[CONFIG_FILE].extend(missing)
+
+        elif service in ('u2f.Authenticate', 'u2f.Register',
```
marmarek, line 172: This is not specific enough - it puts rules for specific U2F keys in the policy file too, but they aren't currently handled correctly by the global config tool: https://github.com/QubesOS/qubes-issues/issues/8525

marmarek (changes requested):
Those are some initial comments. I'll do more tests on some policies I have.

But also, maybe add some tests in this repo too (run it with synthetic input dir and system_info.json file)? it will be easier to test weirder setups this way

qrexec/tools/qrexec_legacy_convert.py
marmarta: this probably means those rules with a specified argument should go into their appropriate file - `60-registered-arguments.policy`

qrexec/tools/qrexec_legacy_convert.py
```diff
@@ -0,0 +1,309 @@
...
+                backup_name = backup_name + '.bak'
+            backups_made[file] = pathlib.Path(backup_name)
+            shutil.copy(file, backup_name)
+
+            if filename != '50-config-input':
+                # input files are special: they replace current rules
+                current_text = file.read_text()
+                if current_text.startswith(TOOL_DISCLAIMER):
+                    current_text = current_text[len(TOOL_DISCLAIMER):]
+                text = text.rstrip('\n') + current_text
```
marmarek, line 267: The rstrip here is wrong at least if the old file doesn't start with an empty line. It will cause last line of converted rules be concatenated with first line of pre-existing rules -> syntax error.

qrexec/tools/qrexec_legacy_convert.py
```diff
@@ -0,0 +1,309 @@
...
+                backup_name = backup_name + '.rpmsave'
+            backups_made[file] = pathlib.Path(backup_name)
+            file.rename(backup_name)
+
+    # check if state changed
+    current_state_string = StringIO()
+    qrexec_policy_graph.main(['--policy-dir',
+                              str(POLICYPATH), '--full-output'],
+                             output=current_state_string)
+    current_state = set(current_state_string.getvalue().split('\n'))
```
marmarek, line 286: Wrap this in some try-except (and in case of error set `current_state` to some value, like empty or "ERROR"). Otherwise, if the tool produce broken policy for some reason (see comment above), it won't offer to rollback.

qrexec/tools/qrexec_legacy_convert.py
```diff
@@ -0,0 +1,309 @@
...
+
+        if self.service == 'qubes.Gpg':
+            if self.rule.source == '@anyvm':
+                if self.rule.target.type == 'keyword':
+                    # not supported main rule
+                    return False
+            return True
+
+        if self.service == 'policy.RegisterArgument':
+            return self.rule.argument == '+u2f.Register'
```
marmarek, line 119:
```suggestion
            return self.rule.argument == '+u2f.Authenticate'
```

marmarek (changes requested):
I've found few more cases that are mishandled:

1. `qubes.Gpg` with `@anyvm @anyvm ask` at the end, but some other rules earlier. For example:
```
work1 keys1 allow
work2 keys2 allow
work2-bis keys2 allow
@anyvm  keys1 deny
@anyvm  keys2 deny
@anyvm  keys3 deny
@anyvm  @anyvm  ask
```
This sets special handling for keys1/2/3 (and some qubes that use them), but otherwise allows user to choose. And more importantly, for interactive selection, those keys1/2/3 are not available (due to the deny rules). The conversion tool puts `@anyvm  @anyvm  ask` into `30-user.policy`, and other rules to `50-config-splitgpg.policy`. And then correctly detects it isn't the same semantics (because in this case order matters). 
This case is important, because default `qubes.Gpg` policy used to have `@anyvm  @anyvm  ask`, so this case will trigger for many users. Maybe put this wildcard rule to `51-config-splitgpg-default.policy` (if found in legacy policy) ? The removal of the wildcard rule in default config was intentional (since we have nice GUI for setting specific rules now), but I don't know how to nicely handle it during conversion, to not risk breaking existing setup (removing this rule might be okay in some cases, but not some others).
Or maybe, just put all `qubes.Gpg` rules into 30-user.policy (preserving their order), and let the user migrate this part manually (see 1a below)?

1a. The global config GUI complains it doesn't understand the "allow" rules for `qubes.Gpg` service from the point above. So, those shouldn't be added to `50-config-splitgpg.policy`.

2. Some updates proxy rules were put in 30-user.policy, instead of 50-config-updates.policy:
```
qubes.UpdatesProxy	*	fedora-39-mintest	@default	allow target=sys-net
qubes.UpdatesProxy	*	@type:TemplateVM	@default	allow target=sys-whonix
```
The first one is expected as custom rule I think, but the second one looks standard, no?

3. `policy.RegisterArgument +u2f.Authenticate` put into `30-user.policy` instead of `50-config-u2f.policy`:
```
policy.RegisterArgument	+u2f.Authenticate	sys-usb	vm1	allow target=@adminvm
policy.RegisterArgument	+u2f.Authenticate	sys-usb	vm2	allow target=@adminvm
policy.RegisterArgument	+u2f.Authenticate	sys-usb	vm3	allow target=@adminvm
policy.RegisterArgument	+u2f.Authenticate	sys-usb	@anyvm	ask default_target=@adminvm
```
Not sure about the last one, but others look standard. This one I found why - see below.

qrexec/tools/qrexec_legacy_convert.py
```diff
@@ -0,0 +1,310 @@
...
+        if self.service == 'qubes.UpdatesProxy':
+            if self.rule.source.type == 'keyword':
+                if str(self.rule.source) not in ['@type:TemplateVM',
+                                                 '@tag:whonix-updatevm']:
+                    return False
+            if 'allow' not in str(self.rule.action):
+                return False
+            return self.rule.target.type != 'keyword' or \
+                str(self.rule.target) == '@adminvm' or \
+                str(self.rule.target == '@default')
```
marmarek, line 102:
```suggestion
                str(self.rule.target) == '@default'
```

marmarek (changes requested):

marmarek (approved):

PR #130, opened 2023-10-24
Fixes to qrexec-policy-graph

marmarek:
Fixes https://github.com/QubesOS/qubes-issues/issues/8648


marmarek:
@ben-grande

ben-grande:
Tested 4002cf9, much better now, thank you.
It fixes showing `qubes.UpdatesProxy` and any other service that have their target as `@default`.

PR #133, opened 2023-11-27
Limit parser target redirect to allowed list

ben-grande:
Limits the parameters target= and default_target= to targets that have no previous deny rule. The previous rule file name and line number is not known and a general information is reported.

Reported-by: unman <unman@thirdeyesecurity.org>
Fixes: https://github.com/QubesOS/qubes-issues/issues/8227

### DIFF
--- a/qrexec/policy/parser.py
+++ b/qrexec/policy/parser.py
@@ -1013,6 +1013,17 @@ class Allow(ActionType):
         elif target == "@adminvm":
             target = "dom0"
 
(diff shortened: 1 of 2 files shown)


ben-grande:
Currently fails because collect_targets_for_ask removes valid values such as `@adminvm` and `@dispvm`.

Waiting for discussion of
- https://github.com/QubesOS/qubes-issues/issues/7723

before continuing.

PR #135, opened 2024-01-15
Implement UUID support in qrexec

DemiMarie:
Open questions:

- [x] How should dom0 be informed of the intended call target?  Should it be `SERVICE-NAME REMOTE-DOMAIN-NAME uuid UUID-OF-TARGET` or `SERVICE-NAME REMOTE-DOMAIN-NAME name @uuid:UUID-OF-TARGET`?
- [ ] Should the target be informed of the source VM’s UUID?  This allows race-free “boomerang” calls, but requires an API change.
- [ ] Probably more
Fixes: QubesOS/qubes-issues#8510

### DIFF
--- a/daemon/qrexec-daemon-common.c
+++ b/daemon/qrexec-daemon-common.c
@@ -11,6 +11,8 @@
 #include "qrexec-daemon-common.h"
 
 const char *socket_dir = QREXEC_DAEMON_SOCKET_DIR;
+#define QREXEC_DISPVM_PREFIX "@dispvm:"
+#define QREXEC_DISPVM_PREFIX_SIZE (sizeof QREXEC_DISPVM_PREFIX - 1)
 
 /* ask the daemon to allocate vchan port */
 bool negotiate_connection_params(int s, int other_domid, unsigned type,
@@ -47,6 +49,20 @@ bool negotiate_connection_params(int s, int other_domid, unsigned type,
     *data_port = params.connect_port;
     *data_domain = params.connect_domain;
     return true;
+}
+
+bool target_refers_to_dom0(const char *target)
+{
+    switch (target[0]) {
+    case '@':
+        return strcmp(target + 1, "adminvm") == 0;
+    case 'd':
+        return strcmp(target + 1, "om0") == 0;
+    case 'u':
+        return strcmp(target + 1, "uid:00000000-0000-0000-0000-000000000000") == 0;
+    default:
+        return false;
+    }
 }
 
 int handle_daemon_handshake(int fd)
@@ -93,7 +109,7 @@ int handle_daemon_handshake(int fd)
 bool qrexec_execute_vm(const char *target, bool autostart, int remote_domain_id,
                        const char *cmd, size_t const service_length,
                        const char *request_id, bool just_exec,
-                       bool wait_connection_end)
+                       bool wait_connection_end, bool use_uuid)
 {
     if (service_length < 2 || (size_t)service_length > MAX_QREXEC_CMD_LEN) {
         /* This is arbitrary, but it helps reduce the risk of overflows in other code */
@@ -119,11 +135,26 @@ bool qrexec_execute_vm(const char *target, bool autostart, int remote_domain_id,
         }
         // Otherwise, we kill the VM immediately after starting it.
         wait_connection_end = true;
-        buf = qubesd_call(target + 8, "admin.vm.CreateDisposable", "", &resp_len);
+        buf = qubesd_call2(target + QREXEC_DISPVM_PREFIX_SIZE,
+                           "admin.vm.CreateDisposable", "",
+                           "uuid", use_uuid ? 4 : 0, &resp_len);
         if (buf == NULL) // error already printed by qubesd_call
             return false;
         if (memcmp(buf, "0", 2) == 0) {
-            /* we exit later so memory leaks do not matter */
+            if (strlen(buf + 2) != resp_len - 2) {
+                LOG(ERROR, "NUL byte in qubesd response");
+                return false;
+            }
+            if (use_uuid) {
+                if (resp_len != sizeof("uuid:00000000-0000-0000-0000-000000000000") + 1) {
+                    LOG(ERROR, "invalid UUID length");
+                    return false;
+                }
+                if (memcmp(buf + 2, "uuid:", 5) != 0) {
+                    LOG(ERROR, "invalid UUID target %s", buf + 2);
+                    return false;
+                }
+            }
             target = buf + 2;
         } else {
             if (memcmp(buf, "2", 2) == 0) {
--- a/daemon/qrexec-daemon.c
+++ b/daemon/qrexec-daemon.c
@@ -20,6 +20,7 @@
  */
 
 #include <inttypes.h>
+#include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/syscall.h>
@@ -135,24 +136,44 @@ static
 #endif
 int protocol_version;
 
-static char *remote_domain_name;	// guess what
+static const char *remote_domain_name;	// guess what
+static const char *remote_domain_uuid;
 static int remote_domain_id;
+
+static void unlink_or_exit(const char *path)
+{
+    int v = unlink(path);
+    if (v != 0 && !(v == -1 && errno == ENOENT))
+        err(1, "unlink(%s)", path);
+}
+
+static char __attribute__((format(printf, 1, 2))) *xasprintf(const char *fmt, ...)
+{
+    va_list x;
+    char *res;
+    va_start(x, fmt);
+    int r = vasprintf(&res, fmt, x);
+    va_end(x);
+    if (r < 0)
+        abort();
+    return res;
+}
 
 static void unlink_qrexec_socket(void)
 {
-    char *socket_address;
-    char *link_to_socket_name;
+    char *socket_name;
+    const char *p[2] = {remote_domain_name, remote_domain_uuid};
+    int i;
 
-    if (asprintf(&socket_address, "%s/qrexec.%d", socket_dir, remote_domain_id) < 0)
-        err(1, "asprintf");
-    if (unlink(socket_address) != 0 && errno != ENOENT)
-        err(1, "unlink(%s)", socket_address);
-    free(socket_address);
-    if (asprintf(&link_to_socket_name, "%s/qrexec.%s", socket_dir, remote_domain_name) < 0)
-        err(1, "asprintf");
-    if (unlink(link_to_socket_name) != 0 && errno != ENOENT)
-        err(1, "unlink(%s)", link_to_socket_name);
-    free(link_to_socket_name);
+    for (i = 0; i < 2; ++i) {
+        char *link_to_socket_name = xasprintf("qrexec.%s%s", i > 0 ? "uuid:" : "", p[i]);
+        unlink_or_exit(link_to_socket_name);
+        free(link_to_socket_name);
+    }
+    if (asprintf(&socket_name, "qrexec.%d", remote_domain_id) < 0)
+        abort();
+    unlink_or_exit(socket_name);
+    free(socket_name);
 }
 
 static void handle_vchan_error(const char *op)
@@ -161,27 +182,25 @@ static void handle_vchan_error(const char *op)
     exit(1);
 }
 
-
-static int create_qrexec_socket(int domid, const char *domname)
+static int create_qrexec_socket(int domid, const char *domname, const char *domuuid)
 {
-    char socket_address[40];
-    char *link_to_socket_name;
-
-    snprintf(socket_address, sizeof(socket_address),
-             "%s/qrexec.%d", socket_dir, domid);
-    if (asprintf(&link_to_socket_name,
-                 "%s/qrexec.%s", socket_dir, domname) < 0)
-        err(1, "asprintf");
-    unlink(link_to_socket_name);
-
     /* When running as root, make the socket accessible; perms on /var/run/qubes still apply */
     umask(0);
-    if (symlink(socket_address, link_to_socket_name)) {
-        PERROR("symlink(%s,%s)", socket_address, link_to_socket_name);
+
+    const char *p[2] = { domuuid, domname };
+    char *socket_address = xasprintf("qrexec.%d", domid);
+    for (int i = 0; i < 2; ++i) {
+        if (p[i] == NULL)
+            continue;
+        char *link_to_socket_name = xasprintf("qrexec.%s%s", i ? "" : "uuid:", p[i]);
+        unlink_or_exit(link_to_socket_name);
+        if (symlink(socket_address, link_to_socket_name)) {
+            PERROR("symlink(%s,%s)", socket_address, link_to_socket_name);
+        }
+        free(link_to_socket_name);
     }
     int fd = get_server_socket(socket_address);
     umask(0077);
-    free(link_to_socket_name);
     return fd;
 }
 
@@ -407,7 +426,7 @@ static void init(int xid, bool opt_direct)
 
     atexit(unlink_qrexec_socket);
     qrexec_daemon_unix_socket_fd =
-        create_qrexec_socket(xid, remote_domain_name);
+        create_qrexec_socket(xid, remote_domain_name, remote_domain_uuid);
 
     struct sigaction sigchld_action = {
         .sa_handler = signal_handler,
@@ -856,11 +875,12 @@ static int parse_policy_response(
     size_t result_bytes,
     bool daemon,
     char **user,
+    char **target_uuid,
     char **target,
     char **requested_target,
     int *autostart
 ) {
-    *user = *target = *requested_target = NULL;
+    *user = *target_uuid = *target = *requested_target = NULL;
     int result = *autostart = -1;
     const char *const msg = daemon ? "qrexec-policy-daemon" : "qrexec-policy-exec";
     // At least one byte must be returned
@@ -904,6 +924,12 @@ static int parse_policy_response(
                 goto bad_response;
             *target = strdup(current_response + (sizeof("target=") - 1));
             if (*target == NULL)
+                abort();
+        } else if (!strncmp(current_response, "target_uuid=", sizeof("target_uuid=") - 1)) {
+            if (*target_uuid != NULL)
+                goto bad_response;
+            *target_uuid = strdup(current_response + 12);
+            if (*target_uuid == NULL)
                 abort();
         } else if (!strncmp(current_response, "autostart=", sizeof("autostart=") - 1)) {
             current_response += sizeof("autostart=") - 1;
@@ -1001,6 +1027,7 @@ static enum policy_response connect_daemon_socket(
         const char *target_domain,
         const char *service_name,
         char **user,
+        char **target_uuid,
         char **target,
         char **requested_target,
         int *autostart
@@ -1028,7 +1055,7 @@ static enum policy_response connect_daemon_socket(
         size_t result_bytes;
         // this closes the socket
         char *result = qubes_read_all_to_malloc(daemon_socket, 64, 4096, &result_bytes);
-        int policy_result = parse_policy_response(result, result_bytes, true, user, target, requested_target, autostart);
+        int policy_result = parse_policy_response(result, result_bytes, true, user, target_uuid, target, requested_target, autostart);
         if (policy_result != RESPONSE_MALFORMED) {
             // This leaks 'result', but as the code execs later anyway this isn't a problem.
             // 'result' cannot be freed as 'user', 'target', and 'requested_target' point into
@@ -1099,7 +1126,7 @@ static enum policy_response connect_daemon_socket(
             // This leaks 'result', but as the code execs later anyway this isn't a problem.
             // 'result' cannot be freed as 'user', 'target', and 'requested_target' point into
             // the same buffer.
-            return parse_policy_response(result, result_bytes, true, user, target, requested_target, autostart);
+            return parse_policy_response(result, result_bytes, true, user, target_uuid, target, requested_target, autostart);
     }
 }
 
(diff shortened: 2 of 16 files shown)


DemiMarie:
PipelineRetryFailed

DemiMarie:
PipelineRetryFailed

marmarek:
>     * [ ]  Should the target be informed of the source VM’s UUID?  This allows race-free “boomerang” calls, but requires an API change.

Definitely not instead of the name, as most(?) services needs actually the source name (things like subdir in QubesIncoming, or split-gpg prompt). As for extra data, its usefulness would also be limited (generally, the source qube is running during the connection, and the name cannot be changed while qube is running). And adding extra info to the protocol would be an API change anyway so needs to wait for R4.3.

DemiMarie:
This version is quite a bit simpler than I thought, but it has no fallback to the name if the UUID symlink is not present, and thus does not work.

Would it be possible to use an environment variable @marmarek?  That could help avoid circular dependencies during upgrade.  Of course, an alternative is to use snapshots to make the entire upgrade atomic.

qrexec/policy/parser.py
```diff
@@ -590,14 +611,13 @@ class DispVMTag(Source, Target):
...
-        other: Optional[str],
+        other: str,
         *,
         system_info: FullSystemInfo,
         source: Optional[VMToken]=None
     ) -> bool:
-        if isinstance(other, DispVM):
-            assert source is not None
-            other = other.get_dispvm_template(source, system_info=system_info)
+        if isinstance(other, DispVM) and source is not None:
```
marmarek, line 619: I'd prefer the assert to stay, instead of silently skipping this branch in that case. Or is None as source valid in this case now for some reason?

qrexec/policy/parser.py
```diff
@@ -698,12 +718,30 @@ async def execute(self) -> str:
...
+target=@dispvm:uuid:{target_info['uuid']}
 autostart={self.autostart}
-requested_target={self.request.target}"""
+requested_target={request.target}"""
+        else:
+            target_info = request.system_info["domains"][target]
+            return f"""\
+user={self.user or 'DEFAULT'}
+result=allow
+target=uuid:{target_info['uuid']}
```
marmarek, line 741: Do not change the meaning of the existing field, it's part of the public API (among other things, it's format of `policy.Eval*` services outout). You can add new `target_uuid` field or something like this.

marmarek (changes requested):

marmarek:
Forcing updating core-admin and core-qrexec at the same time via dependency is okay. There is no need to handle new qubesd starting old qrexec-daemon. Since it isn't necessary, I'd prefer to simply use a command line parameter for UUID, which is much cleaner than giving name via cmdline and UUID via environment.
And also, it's qrexec-daemon who should maintain the UUID symlink, since it maintain the other symlink already (again - consistency).
What might be useful (mostly for testing, as actual release upgrade needs reboot anyway) is to keep working with already running old qrexec-daemon - which basically means a fallback to name-based symlink.

daemon/qrexec-daemon.c
```diff
@@ -838,6 +863,13 @@ static int parse_policy_response(
             *target = strdup(current_response + (sizeof("target=") - 1));
             if (*target == NULL)
                 abort();
+        } else if (!strncmp(current_response, "target_uuid=", sizeof("target_uuid=") - 1)) {
+            if (*target_uuid != NULL)
+                goto bad_response;
+            *target_uuid = strdup(current_response + 7);
+            if (*target_uuid == NULL)
+                abort();
+            (*target_uuid)[4] = ':';
```
marmarek, line 872: This requires a comment what is happening here.

qrexec/policy/parser.py
```diff
@@ -698,12 +716,25 @@ async def execute(self) -> str:
...
 
+        if target in {"@adminvm", "dom0", "uuid:00000000-0000-0000-0000-000000000000"}:
+            return f"""\
+user={self.user or 'DEFAULT'}
+result=allow
+target=@adminvm
+autostart={self.autostart}
+requested_target={request.target}"""
+        if target.startswith("@dispvm:"):
+            target_info = request.system_info["domains"][target[8:]]
```
marmarek, line 727: This looks a bit inconsistent - you need to look at both `target` and `target_uuid` to get full picture. Wouldn't it be better to have `@dispvm:uuid:...` below in this case?

In any case, please document it in `doc/qrexec-policy-daemon.rst`.

daemon/qrexec-daemon.c
```diff
@@ -1102,16 +1143,54 @@ static void handle_execute_service(
...
+        bool disposable = strncmp("@dispvm:", target, sizeof("@dispvm:") - 1) == 0;
+        bool use_uuid = target_uuid != NULL;
+        if (disposable) {
+            char *buf = qubesd_call2(use_uuid ? target_uuid : target + 8,
+                                     "admin.vm.CreateDisposable", "",
+                                     "uuid", use_uuid ? 4 : 0, &resp_len);
             if (!buf) // error already printed by qubesd_call
                 _exit(126);
             if (memcmp(buf, "0", 2) == 0) {
-                /* we exec later so memory leaks do not matter */
```
marmarek, line 1113: This comment is gone, but it's still applicable.

daemon/qrexec-daemon.c
```diff
@@ -1102,16 +1143,54 @@ static void handle_execute_service(
...
+                if (strlen(buf + 2) != resp_len - 2) {
+                    LOG(ERROR, "NUL byte in qubesd response");
+                    _exit(126);
+                }
+                if (use_uuid) {
+                    if (resp_len != sizeof("00000000-0000-0000-0000-000000000000") + 1) {
+                        LOG(ERROR, "invalid UUID length");
+                        _exit(126);
+                    }
+                    target = malloc(resp_len + (sizeof("uuid:") - 2));
```
marmarek, line 1182: Actually, wouldn't it be more consistent if `admin.vm.CreateDisposable` would include `uuid:` prefix in response too?

marmarek (changes requested):

daemon/qrexec-daemon.c
DemiMarie: What is happening is that the response is `target_uuid=SOMEUUID`, but we need `uuid:SOMEUUID`.  That is done by adding 7 (to skip past the `target_`) and then overwriting the `=` with `:`.

daemon/qrexec-daemon.c
DemiMarie: We only exec for VM ⇒ dom0 service calls.  VM ⇒ VM calls do not exec.

daemon/qrexec-daemon.c
DemiMarie: Probably so.  That would allow using the returned string in the Admin API directly.

qrexec/policy/parser.py
DemiMarie: I actually implemented your version and then reverted it.  Your version is better, though, so I will un-revert.

daemon/qrexec-daemon.c
marmarek: Yes, I know. Put that into a comment.

daemon/qrexec-daemon-common.c
```diff
@@ -47,6 +47,20 @@ void negotiate_connection_params(int s, int other_domid, unsigned type,
...
 }
 
+bool target_refers_to_dom0(const char *target)
+{
+    switch (target[0]) {
+    case '@':
+        return strcmp(target + 1, "adminvm") == 0;
+    case 'd':
+        return strcmp(target + 1, "om0") == 0;
+    case 'u':
```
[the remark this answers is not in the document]
DemiMarie: I initially went this route, but it turned out to be quite hard.  `uuid:ZZZ` is really an alternate way to name a VM.

qrexec/tests/policy_parser.py
```diff
@@ -120,6 +134,11 @@ def test_010_Source(self):
         parser.Source("*")
         with self.assertRaises(exc.PolicySyntaxError):
             parser.Source("@default")
+        parser.Source("uuid:d8a249f1-b02b-4944-a9e5-437def2fbe2c")
```
ben-grande, line 137: Please add a test with uppercase letters that don't raise a syntax error. [RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122.html#section-3) allows case insensitive characters on input:

>    Each field is treated as an integer and has its value printed as a
>    zero-filled hexadecimal digit string with the most significant
>    digit first.  The hexadecimal values "a" through "f" are output as
>    lower case characters and are case insensitive on input.

Meaning it is syntactically correct for the user or program to write the uuid value using uppercase letters, but the Qrexec parser should evaluate them with case insensitivity and always output it in lower case.

daemon/qrexec-daemon.c
```diff
@@ -1203,6 +1247,7 @@ _Noreturn static void handle_execute_service_child(
             };
             poll(fds, 1, -1);
             size_t l;
+terminate:
```
marmarek, line 1250: Do not jump into some nested control structure. https://xkcd.com/292/
I sometimes tolerate goto to do cleanups at the end of the function (since C has no try/finally or similar constructs), but do not abuse it.
(and in fact I think some jumps into here are wrong, may kill non-disposable qube too).

daemon/qrexec-daemon.c
```diff
@@ -1139,15 +1159,24 @@ _Noreturn static void handle_execute_service_child(
...
+        bool const disposable = strncmp(QREXEC_DISPVM_PREFIX, selected_target, QREXEC_DISPVM_PREFIX_SIZE) == 0;
+        if (disposable) {
+            char *buf = qubesd_call2(selected_target + QREXEC_DISPVM_PREFIX_SIZE,
+                                     "admin.vm.CreateDisposable", "",
+                                     "uuid", use_uuid ? 4 : 0, &resp_len);
             if (!buf) // error already printed by qubesd_call
                 daemon__exit(126);
             if (memcmp(buf, "0", 2) == 0) {
-                /* we exec later so memory leaks do not matter */
-                target = buf + 2;
```
marmarek, line 1150: After last force-push the target isn't updated with name/uuid of the just started DispVM anymore.

marmarek:
This got some conflicts in the meantime :(

marmarek:
This doesn't work just after upgrade either...
After updating dom0, templates update failed. Journalctl says:
```
Jun 28 19:09:14 dom0 qrexec-policy-daemon[26492]: qrexec: qubes.UpdatesProxy+: debian-12-xfce -> @default: allowed to sys-net
Jun 28 19:09:14 dom0 qrexec-policy-daemon[26492]: qrexec: qubes.UpdatesProxy+: debian-12-xfce -> @default: allowed to sys-net
```
but qrexec log of the template says:
```
2024-06-28 19:09:14.801 qrexec-daemon[28842]: qrexec-daemon-common.c:263:connect_unix_socket: connect /var/run/qubes/qrexec.uuid:78fb327b-f31b-4b33-98f2-8c68ea7009ba
2024-06-28 19:09:14.806 qrexec-daemon[28842]: qrexec-daemon-common.c:263:connect_unix_socket: connect /var/run/qubes/qrexec.uuid:78fb327b-f31b-4b33-98f2-8c68ea7009ba
```
and since sys-net was started before the update, it fails. BTW, the message should include actual error...

Any ideas how to solve it in elegant way? IIUC adding a fallback to name-based symlink will defeat the purpose of this change. I guess the upgrade tool must ensure all VMs are restarted after updating dom0. Any better idea?

DemiMarie:
I think the simplest option is to just force a reboot after upgrading dom0.

DemiMarie:
> Any ideas how to solve it in elegant way? IIUC adding a fallback to name-based symlink will defeat the purpose of this change. I guess the upgrade tool must ensure all VMs are restarted after updating dom0. Any better idea?

Don’t we need to reboot dom0 anyway due to a Xen/toolstack version mismatch?

marmarek:
> system_tests_gui_interactive
> 
>     * gui_filecopy: [unnamed test](https://openqa.qubes-os.org/tests/104605#step/gui_filecopy/18) (unknown)
> 
>     * gui_filecopy: [Failed](https://openqa.qubes-os.org/tests/104605#step/gui_filecopy/19) (test died)
>       `# Test died: no candidate needle with tag(s) 'file-copy-prompt' mat...`

File copy was denied before showing a prompt. Journalctl shows:
```
Jun 28 23:45:27 dom0 qrexec-policy-daemon[1427]: invalid ask response for qubes.Filecopy:
Jun 28 23:45:27 dom0 qrexec-policy-daemon[1427]: qrexec: qubes.Filecopy: work -> @default: denied: invalid response
```
I don't see any log message from qrexec-policy-agent, but I guess it didn't like something about this change.
In this case, whole system was restarted after update, so it isn't upgrade path issue.

DemiMarie:
I’ll do some more testing.  I don’t think I tested `ask`, only `allow` and possibly `deny`.

marmarek:
ping?

DemiMarie:
> ping?

got distracted dealing with Wayland stuff, sorry :laughing:.

qrexec/policy/parser.py
```diff
@@ -1651,10 +1693,17 @@ def collect_targets_for_ask(self, request):
...
-                    rule_target.expand(system_info=request.system_info)
-                )
-
+                expansion = set()
+                for potential_target in rule_target.expand(
+                        system_info=request.system_info):
+                    try:
+                        # The policy agent cannot handle UUIDs (and rightly so,
+                        # those are meaningless to humans). Convert them to names.
+                        # FIXME: this causes dispvm name reuse issues, should
```
marmarek, line 1702: First, does it really? You later translate name back to uuid using `request.system_info`, which was fetched at the request (not agent response) time. But second, now dispvm names are not reused for at least a week, so this comment doesn't really apply.

DemiMarie: Fixed.

qrexec/policy/parser.py
```diff
@@ -1651,10 +1693,19 @@ def collect_targets_for_ask(self, request):
...
+                expansion = set()
+                for potential_target in rule_target.expand(
+                        system_info=request.system_info):
+                    try:
+                        # The policy agent cannot handle UUIDs (and rightly so,
+                        # those are meaningless to humans). Convert them to names.
+                        # VM names can be reused, but in practice, this is unlikely
+                        # to happen except by user request or DispVM name reuse.
+                        # The latter will not happen for a week and so is unlikely
+                        # to confuse humans.
```
marmarek, line 1705: I think you ignored first sentence of my earlier comment. Even if VM name gets reused, you translate policy agent response back to UUID using dict collected at the request time. So, if VM with that  **UUID** doesn't exist anymore, the call will fail, it won't make it to the same-named new VM. DispVMs are not any different here.
Have I missed anything?

BTW, it might be useful to add yet another dict to the agent params, with name->uuid mapping of the possible targets (+ source too?), just in case some future/alternative agent would like to use it. But due to above, other parts can still use names, so it won't break compatibility with old agent.

marmarek (changes requested):
Still doesn't work:
```
Task exception was never retrieved
future: <Task finished name='Task-4' coro=<SocketService._client_connected() done, defined at /usr/lib/python3.11/site-packages/qrexec/server.py:54> exception=ValueError('Input string was expected to be sanitized, but was not.')>
  File "/usr/lib/python3.11/site-packages/qrexec/server.py", line 71, in _client_connected
    response = await self.handle_request(params, service, source_domain)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_agent.py", line 557, in handle_request
    return await self.handle_ask(params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_agent.py", line 574, in handle_ask
    target = await confirm_rpc(
             ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_agent.py", line 525, in confirm_rpc
    window = RPCConfirmationWindow(
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_agent.py", line 478, in __init__
    list_modeler = self._new_vm_list_modeler()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_agent.py", line 507, in _new_vm_list_modeler
    return VMListModeler(self._entries_info)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_agent.py", line 58, in __init__
    self._create_entries()
  File "/usr/lib/python3.11/site-packages/qrexec/tools/qrexec_policy_agent.py", line 79, in _create_entries
    sanitize_domain_name(vm_name, assert_sanitized=True)
  File "/usr/lib/python3.11/site-packages/qrexec/utils.py", line 71, in sanitize_domain_name
    return _sanitize_name(input_string, set(), assert_sanitized)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/qrexec/utils.py", line 64, in _sanitize_name
    raise ValueError(
ValueError: Input string was expected to be sanitized, but was not.
```

marmarek:
The change you made should be detected by tests. I think `qrexec/tests/cli.py` needs an update to `MockPolicy` object to include UUID changes.

marmarek (changes requested):

qrexec/policy/parser.py
```diff
@@ -297,21 +312,38 @@ def __init__(self, token: str, *, filepath: Optional[pathlib.Path]=None,
...
 
+    def __hash__(self):
+        return super().__hash__()
+
+    def __eq__(self, other):
+        """
+        Trap if we ever try to compare a UUID with a non-UUID.
+        This indicates a bug, so throw an exception.
+        """
+        if ((type(other) is not str) and
```
marmarek, line 323: Pylint (rightfully) complains about this. I don't think you even needs this check.
Anyway, you try very hard to break `__eq__` operator on uuid. Why?
There is `match` method that is used for comparing if given policy token matches given VM (source, target etc). `__eq__` operator is just that - comparing tokens on their own, not their meaning (in some cases it used to be the same, and in those cases the `match()` method used normal comparison, but this PR is supposed to update relevant matching methods). There is no expectation that `"@type:template" == "fedora-40"` or even `VMToken("@type:template") == VMToken("fedora-40")`.

marmarek (changes requested):
In addition to previous not addressed comments:

qrexec/policy/parser.py
marmarek: As said above: the `==` operator is _not_ supposed to answer whether given tokens describe the same object. That's what `match()` method is for. If there are places that use `==` for that, they need to be fixed to use the `match()` method. I see a few places that checks for explicit `@adminvm` or `dom0`, and at least one of them actually mean literal `dom0`, not any other representation of that token.
From what I see, (almost?) all your changes to avoid using this operator are _only_ to avoid hitting this exception. From what I see, the _only_ place were `==` usage needed to be changed was automatic conversion of `dom0` (and now `uuid:000...000`) to `@adminvm`. And yes, this single place needs a change.

Instrumenting `__eq__` might be valid development strategy, but it has no place in the final code.

qrexec/policy/parser.py
```diff
@@ -208,6 +208,20 @@ def __init__(cls, name: str, bases: Tuple[type], dict_: Dict[str, str]):
         if "PREFIX" in dict_:
             cls.prefixes[dict_["PREFIX"]] = cls # type: ignore
 
+def match_strings(info: SystemInfo, self: str, other: str) -> bool:
```
marmarek, line 211: Shouldn't this be a method of `VMToken` class?

DemiMarie: I believe it can be called with arguments that are not `VMToken`s, and the operation is symmetric between `self` and `other`.

qrexec/policy/parser.py
```diff
@@ -233,14 +247,15 @@ class VMToken(str, metaclass=VMTokenMeta):
     def __new__(cls, token: str, *, filepath: Optional[pathlib.Path]=None,
                 lineno: Optional[int]=None) -> "VMToken":
         orig_token = token
+        str_eq = str.__eq__
 
         # first, adjust some aliases
-        if token == "dom0":
+        if str_eq(token, "dom0") or str_eq(token, "uuid:00000000-0000-0000-0000-000000000000"):
```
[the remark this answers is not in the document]
DemiMarie: `__eq__` is no longer overridden.

qrexec/policy/parser.py
DemiMarie: Instrumentation removed.

qrexec/policy/parser.py
```diff
@@ -297,21 +312,38 @@ def __init__(self, token: str, *, filepath: Optional[pathlib.Path]=None,
...
-        return self == other
+        # pylint: disable=unused-argument,too-many-return-statements
+        if self == "@adminvm":
+            return other == "@adminvm"
+        return match_strings(system_info["domains"], self, other)
 
     def is_special_value(self) -> bool:
         """Check if the token specification is special (keyword) value"""
-        return self.startswith("@") or self == "*"
+        return self.startswith("@") or "*" == self
```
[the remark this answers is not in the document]
DemiMarie: Reverted.

marmarek:
pylint complains (minor thing)

qrexec/tests/policy_parser.py
ben-grande: I agree on having a strict syntax check for lower case letters.

PR #136, opened 2024-02-10
Avoid using qrexec-client for VM ⇒ VM calls

DemiMarie:
This saves an `execve()`.  It also allows qrexec-daemon to change how it performs a VM -> VM service call without having to change qrexec-client.  During an upgrade, it is possible that qrexec-daemon is newer than qrexec-client, causing the new qrexec-daemon to try to use a calling convention that the old qrexec-client doesn't support.  Doing VM -> VM calls without calling execve() means this cannot happen.  VM -> dom0 and dom0 -> VM calls still use qrexec-client, but VM -> dom0 calls are safe from domain name reuse races as of 654635c2a374272f18bdfe59472a8ae6b51a551e, and the latter do not involve qrexec-daemon at all.

The preceding commits are all either cleanups or bug fixes.

### DIFF
--- a/daemon/qrexec-daemon-common.c
+++ b/daemon/qrexec-daemon-common.c
@@ -1,0 +1,152 @@
+#include <stdlib.h>
+#include <assert.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <stdio.h>
+
+#include "qrexec.h"
+#include "libqrexec-utils.h"
+#include "qrexec-daemon-common.h"
+
+const char *socket_dir = QREXEC_DAEMON_SOCKET_DIR;
+
+/* ask the daemon to allocate vchan port */
+bool negotiate_connection_params(int s, int other_domid, unsigned type,
+        void *cmdline_param, int cmdline_size,
+        int *data_domain, int *data_port)
+{
+    struct msg_header hdr;
+    struct exec_params params;
+    hdr.type = type;
+    hdr.len = sizeof(params) + cmdline_size;
+    params.connect_domain = other_domid;
+    params.connect_port = 0;
+    if (!write_all(s, &hdr, sizeof(hdr))
+            || !write_all(s, &params, sizeof(params))
+            || !write_all(s, cmdline_param, cmdline_size)) {
+        PERROR("write daemon");
+        return false;
+    }
+    /* the daemon will respond with the same message with connect_port filled
+     * and empty cmdline */
+    if (!read_all(s, &hdr, sizeof(hdr))) {
+        PERROR("read daemon");
+        return false;
+    }
+    assert(hdr.type == type);
+    if (hdr.len != sizeof(params)) {
+        LOG(ERROR, "Invalid response for 0x%x", type);
+        return false;
+    }
+    if (!read_all(s, &params, sizeof(params))) {
+        PERROR("read daemon");
+        return false;
+    }
+    *data_port = params.connect_port;
+    *data_domain = params.connect_domain;
+    return true;
+}
+
+int handle_daemon_handshake(int fd)
+{
+    struct msg_header hdr;
+    struct peer_info info;
+
+    /* daemon send MSG_HELLO first */
+    if (!read_all(fd, &hdr, sizeof(hdr))) {
+        PERROR("daemon handshake");
+        return -1;
+    }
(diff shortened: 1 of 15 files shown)


marmarek (changes requested):

marmarek:
Please take a look at test coverage and verify that correct code is used. I'd expected functions like `handle_agent_handshake()`, `select_loop()` or `run_qrexec_to_dom0()` to be covered by tests already.

DemiMarie:
Culprit is missing `__gcov_flush()` calls.

> Please take a look at test coverage and verify that correct code is used. I'd expected functions like `handle_agent_handshake()`, `select_loop()` or `run_qrexec_to_dom0()` to be covered by tests already.

I added `abort()` calls at the start of these functions and the tests fail as expected.

marmarek:
This needs a rebase...

DemiMarie:
…which I was already working on.

marmarek:
> During an upgrade, it is possible that qrexec-daemon is newer than
qrexec-client, causing the new qrexec-daemon to try to use a calling
convention that the old qrexec-client doesn't support.

I think you meant the other way around (newer qrexec-client)

DemiMarie:
> > During an upgrade, it is possible that qrexec-daemon is newer than
> > qrexec-client, causing the new qrexec-daemon to try to use a calling
> > convention that the old qrexec-client doesn't support.
> 
> I think you meant the other way around (newer qrexec-client)

I probably meant what I wrote, but what I meant was incorrect.  Will fix.

daemon/qrexec-daemon-common.c
```diff
@@ -0,0 +1,410 @@
...
+                        int connection_timeout,
+                        bool exit_with_code)
+{
+    int data_domain;
+    int data_port;
+    int s;
+    libvchan_t *data_vchan = NULL;
+
+    set_remote_domain(src_domain_name);
+    s = connect_unix_socket_by_id(src_domain_id);
```
marmarek, line 355: Missing error handling

marmarek (changes requested):
Besides a comment below, please test old qrexec-daemon + new qrexec-client still works (update dom0 package while some VM is running, and then make call to dom0 from that VM). I don't see any issue for this case, but better double check.

daemon/qrexec-daemon-common.c
DemiMarie: Fixed.

DemiMarie:
> Besides a comment below, please test old qrexec-daemon + new qrexec-client still works (update dom0 package while some VM is running, and then make call to dom0 from that VM). I don't see any issue for this case, but better double check.

Tested with VM ⇒ dom0, VM ⇒ VM, and dom0 ⇒ VM service calls, both before and after the change.  All work fine, both with new and existing qrexec-daemon instances.

marmarek (approved):

DemiMarie:
All but the last commit is safe to merge.  The last commit has a problem: if there are certain errors (such as `fork()` failing, which can happen!), the socket will be deleted.  I’ll fix this.

DemiMarie:
This sort of breaks the tests: the test suite passes (at least on my machine), but qrexec-client is only used to _test_ calls to dom0, not to actually execute them.  Most of the code (`run_qrexec_to_dom0()`) is shared between qrexec-daemon and qrexec-client, though.

marmarek:
> @DemiMarie
[Avoid using alarm(2) for timeouts](https://github.com/QubesOS/qubes-core-qrexec/pull/136/commits/8bea1f179df1c92f0d2a8fa034f92e4b53b8abf1)

Please, don't put new changes into already approved PR. Move it to another PR.

And also, prefer short commit ids for "Fixes" tag (try to keep commit messages not too wide).

DemiMarie:
> > @DemiMarie
> > [Avoid using alarm(2) for timeouts](https://github.com/QubesOS/qubes-core-qrexec/pull/136/commits/8bea1f179df1c92f0d2a8fa034f92e4b53b8abf1)
> 
> Please, don't put new changes into already approved PR. Move it to another PR.

Sorry, I thought this PR was no longer approved because I had to force-push it to fix a bug.  I will drop this patch and make a separate PR on top of this one.  (If it is separate, merge conflicts galore.)

> And also, prefer short commit ids for "Fixes" tag (try to keep commit messages not too wide).

Sure.

marmarek (approved):

PR #138, opened 2024-03-23
Support not passing metadata to socket-based services

DemiMarie:
This makes it easier to implement socket-based services that do not
require the metadata.  This avoids having to use a slow executable-based
service or write a custom wrapper.

Fixes: https://github.com/QubesOS/qubes-issues/issues/9036

Edit: This PR now also fixes https://github.com/QubesOS/qubes-issues/issues/9073.  Furthermore, the first four commits fix multiple bugs in the test suite, and the fifth and sixth are purely test suite enhancements.

### DIFF
--- a/qrexec/tests/socket/agent.py
+++ b/qrexec/tests/socket/agent.py
@@ -304,7 +304,16 @@ class TestAgentExecQubesRpc(TestAgentBase):
 
         target = self.connect_target()
         target.handshake()
-        return target
+        return target, dom0
+
+    def check_dom0(self, dom0):
+        self.assertEqual(
+            dom0.recv_message(),
+            (
+                qrexec.MSG_CONNECTION_TERMINATED,
+                struct.pack("<LL", self.target_domain, self.target_port),
+            ),
+        )
 
     def make_executable_service(self, *args):
         util.make_executable_service(self.tempdir, *args)
@@ -319,7 +328,7 @@ class TestAgentExecQubesRpc(TestAgentBase):
 echo "arg: $1, remote domain: $QREXEC_REMOTE_DOMAIN"
 """,
         )
-        target = self.execute_qubesrpc("qubes.Service+arg", "domX")
+        target, _ = self.execute_qubesrpc("qubes.Service+arg", "domX")
         target.send_message(qrexec.MSG_DATA_STDIN, b"")
         messages = target.recv_all_messages()
         self.assertListEqual(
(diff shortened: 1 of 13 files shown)


marmarek:
There are unit tests for socket services (look for `qubes.SocketService`), can you extend them to check this new option too?

DemiMarie:
> There are unit tests for socket services (look for `qubes.SocketService`), can you extend them to check this new option too?

Added.  They caught a nasty bug in the TOML code as well.

qrexec/tests/socket/agent.py
```diff
@@ -438,6 +438,41 @@ def test_exec_service_with_arg(self):
...
+        self.addCleanup(server.close)
+
+        target = self.execute_qubesrpc("qubes.SocketService+arg2", "domX")
+
+        server.accept()
+
+        message = b"stdin data"
+        target.send_message(qrexec.MSG_DATA_STDIN, message)
+        target.send_message(qrexec.MSG_DATA_STDIN, b"")
+        self.assertEqual(server.recvall(len(message) - 1), message[:-1])
```
marmarek, line 461: Why -1?

DemiMarie: I was trying to figure out why the test was hanging.  It turns out that it was failing to load the config file.

marmarek:
PipelineRetry

marmarek:
PipelineRetry

DemiMarie:
> Finally, we have test coverage reporting here :)
> 
> I see it claims some parts of `load_service_config_v2()` are not covered, that I think do have some tests already (especially setting wait-for-session=1), can you verify it? (maybe it's another case where test-provided config isn't actually used? or maybe it's a legitimate code issue?)
> 
> And also, I see some low hanging fruits regarding toml parser:
> 
> * string option (add any test setting `user='user'`, even though actual user switching isn't supported in unit tests)
> * false boolean value, integer starting with 0

Right now, it’s very hard to test (or debug!) services that either invalid command lines or invalid configs.  The connection simply hangs with no error message.  I could add a hacky workaround, but I’d prefer a proper solution: have the agent send `MSG_SERVICE_REFUSED` in this case.  I believe @omeg will be working on this.

marmarek:
> have the agent send `MSG_SERVICE_REFUSED` in this case

No, that wont work, as this is only daemon->agent message (for a reason). At this stage, the calling qube already tries to establish data vchan, and it needs to get a failure message somehow (or wait until timeout, which is less helpful). So, the proper error handling at this stage is to establish that data vchan and send error as an exit code. Sending also an error message on stderr might be an option, but I'd be careful about that, as it might be confusing.

DemiMarie:
Now fixed in f7ec4b0f67f2f730f639dcda249310cae647294c.  Additionally, 518ec17757092a6be58f6d8c31a9789939db7995 adds support for running tests with sanitizers and 07e2c3c31fa8d118af606e64e06c0f791278e5cb fixes what I believe is a race condition in the test suite.  I had to use clang for ASAN+UBSAN builds, as binaries built with GCC crashed with SIGSEGV.

marmarek (approved):

DemiMarie:
This has a bug: skip-service-descriptor is _not_ honored for calls to dom0.  #136 fixes this.

PR #139, opened 2024-04-05
Various bugfixes

DemiMarie:
The first six commits fix user-visible bugs.  The last two fix a bug that is currently unobservable, but the code is fragile and I want to make sure it doesn’t break in the future.

Each of the fixes but the last is split into two commits.  The first commit adds a unit test marked as `@unittest.expectedFailure`.  The second commit is the actual fix and removes the `@unittest.expectedFailure` decorator.  This allowed me to ensure that the tests failed before the bug was fixed, and passed afterwards.

Fixes: QubesOS/qubes-issues#9089
Fixes: QubesOS/qubes-issues#9090
Fixes: QubesOS/qubes-issues#9091
Fixes: QubesOS/qubes-issues#9097
Fixes: QubesOS/qubes-issues#9098
Fixes: QubesOS/qubes-issues#9099
Fixes: QubesOS/qubes-issues#9100
Fixes: QubesOS/qubes-issues#9101


daemon/qrexec-daemon.c
```diff
@@ -1262,6 +1262,23 @@ static bool validate_request_id(struct service_params *untrusted_params, const c
...
 #define ENSURE_NULL_TERMINATED(x) x[sizeof(x)-1] = 0
 
+static bool validate_service_name(char *untrusted_service_name)
+{
+    switch (untrusted_service_name[0]) {
+    case '\0':
+        LOG(ERROR, "Empty service name not allowed");
+        return false;
+    case '+':
+    // TODO: should '-' be blocked, both here and in the first character of the argument?
```
marmarek, line 1272: No, it shouldn't, it's a valid character to have here, and I think it even has some users (yes, also as the first character of the argument).

DemiMarie: That’s unfortunate.  I was hoping to block it so that e.g. a symlink to `/bin/cat` would be a valid echo service.  (Otherwise, there is an option injection bug, though I am not sure if it is exploitable there.)  Anyway, I deleted the comment.

libqrexec/exec.c
```diff
@@ -264,7 +282,7 @@ static int load_service_config_raw(struct qrexec_parsed_command *cmd,
 
     int ret = find_file(config_path, cmd->service_descriptor,
                         config_full_path, sizeof(config_full_path), NULL);
-    if (ret < 0 && strcmp(cmd->service_descriptor, cmd->service_name) != 0)
+    if (ret == -1 && cmd->service_descriptor != cmd->service_name)
```
marmarek, line 285: Why? I don't think your new code makes them point at the same buffer ever. But also, this looks like assumption the strings share the buffer if they are the same, which isn't even true.

libqrexec/exec.c
```diff
@@ -343,53 +361,61 @@ struct qrexec_parsed_command *parse_qubes_rpc_command(
...
         end = strchrnul(start, ' ');
-        cmd->source_domain = strndup(start, end - start);
+        size_t const source_domain_len = (size_t)(end - start);
+        cmd->source_domain = malloc(source_domain_len + 1);
         if (!cmd->source_domain) {
             PERROR("strndup");
             goto err;
         }
+        memcpy(cmd->source_domain, start, source_domain_len);
+        cmd->source_domain[source_domain_len] = '\0';
```
marmarek, line 418: Why open-code strndup?

libqrexec/exec.c
```diff
@@ -343,53 +361,61 @@ struct qrexec_parsed_command *parse_qubes_rpc_command(
...
+            LOG(ERROR, "Service name empty");
+            goto err;
+        }
+        cmd->service_name = malloc(name_len + 1);
+        if (!cmd->service_name) {
+            PERROR("malloc");
+            goto err;
+        }
+        memcpy(cmd->service_name, start, name_len);
+        cmd->service_name[name_len] = '\0';
```
marmarek, line 393: Why open-code strndup?

libqrexec/exec.c
```diff
@@ -343,53 +361,61 @@ struct qrexec_parsed_command *parse_qubes_rpc_command(
...
+        size_t const descriptor_buffer_size = descriptor_len + 1 + (plus == NULL);
+        cmd->service_descriptor = malloc(descriptor_buffer_size);
+        if (!cmd->service_descriptor) {
+            PERROR("malloc");
+            goto err;
         }
+        memcpy(cmd->service_descriptor, start, descriptor_len);
+        /* if plus != NULL, this + will be immediately overwritten */
+        cmd->service_descriptor[descriptor_len] = '+';
+        cmd->service_descriptor[descriptor_buffer_size - 1] = '\0';
```
marmarek, line 405: This could remain strndup + one extra line that adds `+` if necessary.

libqrexec/exec.c
```diff
@@ -236,15 +242,27 @@ static int find_file(
 
         if (path_length + name_length + 1 >= buffer_size) {
             LOG(ERROR, "find_qrexec_service_file: buffer too small for file path");
-            return -1;
+            return -2;
```
marmarek, line 245: Too long argument for a file path should still continue the search for service file without argument.

marmarek (changes requested):

libqrexec/exec.c
DemiMarie: This should never happen with the default search path.  Names longer than `NAME_MAX` are rejected earlier: https://github.com/QubesOS/qubes-core-qrexec/blob/6787a57af73558f949b33a6479d0ba451d078593/libqrexec/exec.c#L232-L233

So this really means that either you have a really long entry in the search path, or there is a bug somewhere.

libqrexec/exec.c
DemiMarie: This is a relic of an earlier draft that _did_ make this guarantee.  Now fixed.

libqrexec/exec.c
DemiMarie: `strndup()` has an implicit `strnlen()` (to get the buffer size) which isn’t needed here (dom0 promises there are no intermediate NUL bytes).  I added a `memdupnul()` function that also has a `PERROR` on failure.

libqrexec/exec.c
DemiMarie: See above.

libqrexec/exec.c
DemiMarie: See above.  The branchless trick (exploiting pointer aliasing) is cute, but I got rid of it.  Also, it takes a little bit of thought to realize that `strndup()` won’t allocate too little memory.  (It won’t because of the space after the service descriptor.)

libqrexec/exec.c
marmarek: What about service+arg being for example NAME_MAX-1 long?

DemiMarie (commented):
This led to three more bugs being discovered and fixed.

marmarek (approved):

PR #141, opened 2024-04-10
Stop using qubes-rpc-multiplexer

DemiMarie:
Instead, directly execute the command from C.

The command’s environment inherits all environment variables from the
calling process _except_ for those that start with `QREXEC`.  Environment
variables that start with `QREXEC` are _not_ inherited, except for
`QREXEC_SERVICE_PATH` and `QREXEC_AGENT_PID`, but the following are set
explicitly:

- `QREXEC_SERVICE_FULL_NAME`
- `QREXEC_REMOTE_DOMAIN`
- `QREXEC_SERVICE_ARGUMENT` if the service argument is present and non-empty.
- `QREXEC_AGENT_PID` is set for calls to a VM that do not use `MSG_JUST_EXEC`.
  Its handling in other cases is inconsistent, so such services should avoid it.

When the request is made to dom0, the following are also set:

- `QREXEC_REQUESTED_TARGET_TYPE` is set to `name` if the requested target does not start with
- `QREXEC_REQUESTED_TARGET` for services in dom0 if the target does not start with `@`.
- `QREXEC_REQUESTED_TARGET_KEYWORD` for services in dom0 if the target does start with `@`.
  The leading `@` is removed, but subsequent `@` are left unchanged.

This is a backwards-incompatible change to libqrexec.  Therefore, it cannot be backported to R4.2.

Fixes: QubesOS/qubes-issues#9062

### DIFF
--- a/libqrexec/exec.c
+++ b/libqrexec/exec.c
@@ -786,7 +786,9 @@ int find_qrexec_service(
     const char *qrexec_service_path = getenv("QREXEC_SERVICE_PATH");
     if (!qrexec_service_path)
         qrexec_service_path = QREXEC_SERVICE_PATH;
-    *socket_fd = -1;
+    if (socket_fd != NULL) {
+        *socket_fd = -1;
+    }
 
     struct stat statbuf;
 
@@ -807,6 +809,9 @@ int find_qrexec_service(
 
     if (S_ISSOCK(statbuf.st_mode)) {
         /* Socket-based service. */
+        if (socket_fd == NULL) {
+            goto bad_socket_service;
+        }
         int s;
         if ((s = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0)) == -1) {
             PERROR("socket");
@@ -827,6 +832,9 @@ int find_qrexec_service(
         *socket_fd = s;
         return 0;
     } else if (S_ISLNK(statbuf.st_mode)) {
+        if (socket_fd == NULL) {
+            goto bad_socket_service;
+        }
         /* TCP-based service */
         assert(path_buffer->buflen >= (int)sizeof("/dev/tcp") - 1);
         assert(memcmp(path_buffer->data, "/dev/tcp", sizeof("/dev/tcp") - 1) == 0);
@@ -914,6 +922,11 @@ int find_qrexec_service(
     }
 
     LOG(ERROR, "Unknown service type (not executable, not a socket): %.*s",
+        path_buffer->buflen, path_buffer->data);
+    return -2;
+bad_socket_service:
+    LOG(ERROR, "Service %.*s is a socket or /dev/tcp symlink, but this is not "
+               "supported for dom0 -> dom0 calls",
         path_buffer->buflen, path_buffer->data);
     return -2;
 }
--- a/libqrexec/exec.c
+++ b/libqrexec/exec.c
@@ -47,43 +47,180 @@ void register_exec_func(do_exec_t *func) {
     exec_func = func;
 }
 
-void exec_qubes_rpc_if_requested(const char *prog, char *const envp[]) {
-    /* avoid calling qubes-rpc-multiplexer through shell */
-    if (strncmp(prog, RPC_REQUEST_COMMAND, RPC_REQUEST_COMMAND_LEN) == 0) {
+#define STR_LITERAL_SIZE(a) (sizeof("" a) - 1)
+#define STARTS_WITH_LITERAL(a, b) (strncmp(a, b, STR_LITERAL_SIZE(b)) == 0)
+
+static bool should_strip_env_var(const char *var)
+{
+    if (!STARTS_WITH_LITERAL(var, "QREXEC"))
+        return false;
+    const char *rest = var + STR_LITERAL_SIZE("QREXEC");
+    return !STARTS_WITH_LITERAL(rest, "_SERVICE_PATH=") &&
+           !STARTS_WITH_LITERAL(rest, "_AGENT_PID=");
+}
+
+void exec_qubes_rpc_if_requested2(const char *program, const char *cmd, char *const envp[],
+                                  bool use_shell) {
+    /* avoid calling RPC service through shell */
+    if (program) {
+        assert(program);
         char *prog_copy;
         char *tok, *savetok;
-        char *argv[16]; // right now 6 are used, but allow future extensions
-        size_t i = 1;
+        const char *argv[10];
+        size_t i = 0;
+#define MAX_ADDED_ENV_VARS 5
+        size_t const extra_env_vars = MAX_ADDED_ENV_VARS;
+        size_t env_amount = extra_env_vars;
 
-        if (prog[RPC_REQUEST_COMMAND_LEN] != ' ') {
-            LOG(ERROR, "\"" RPC_REQUEST_COMMAND "\" not followed by space");
-            _exit(126);
+        if (strncmp(cmd, RPC_REQUEST_COMMAND " ", RPC_REQUEST_COMMAND_LEN + 1) != 0) {
+            LOG(ERROR, "program != NULL, but '%s' does not start with '%s '",
+                cmd, RPC_REQUEST_COMMAND " ");
+            assert(!"Invalid command");
+            _exit(QREXEC_EXIT_PROBLEM);
         }
 
-        prog_copy = strdup(prog + RPC_REQUEST_COMMAND_LEN + 1);
+        for (char *const *env = envp; *env; ++env) {
+            // Set this 0 to 1 if adding new variable settings below,
+            // to ensure that MAX_ADDED_ENV_VARS is correct.
+            if (0 && should_strip_env_var(*env))
+                continue;
+            env_amount++;
+        }
+#define EXTEND(...)                                             \
+        do {                                                    \
+            if (iterator >= env_amount)                         \
+                abort();                                        \
+            if (asprintf(&buf[iterator++], __VA_ARGS__) < 0)    \
+                goto bad_asprintf;                              \
+        } while (0)
+#define EXTEND_RAW(arg)                                         \
+        do {                                                    \
+            if (iterator >= env_amount)                         \
+                abort();                                        \
+            buf[iterator++] = (arg);                            \
+        } while (0)
+
+        char **buf = calloc(env_amount + 1, sizeof(char *));
+        if (buf == NULL) {
+            LOG(ERROR, "calloc(%zu, %zu) failed: %m", env_amount, sizeof(char *));
+            _exit(QREXEC_EXIT_PROBLEM);
+        }
+        size_t iterator = 0;
+        for (char *const *env = envp; *env; ++env) {
+            if (!should_strip_env_var(*env)) {
+                EXTEND_RAW(*env);
+            }
+        }
+
+        prog_copy = strdup(cmd + RPC_REQUEST_COMMAND_LEN + 1);
         if (!prog_copy) {
             PERROR("strdup");
             _exit(QREXEC_EXIT_PROBLEM);
         }
 
+        argv[i++] = program;
         tok=strtok_r(prog_copy, " ", &savetok);
         while (tok != NULL) {
-            if (i >= sizeof(argv)/sizeof(argv[0])-1) {
-                LOG(ERROR, "To many arguments to %s", RPC_REQUEST_COMMAND);
+            if (i >= sizeof(argv)/sizeof(argv[0])) {
+                LOG(ERROR, "Too many arguments to %s", RPC_REQUEST_COMMAND);
                 _exit(QREXEC_EXIT_PROBLEM);
             }
             argv[i++] = tok;
-            tok=strtok_r(NULL, " ", &savetok);
+            tok = strtok_r(NULL, " ", &savetok);
         }
-        argv[i] = NULL;
 
-        argv[0] = getenv("QREXEC_MULTIPLEXER_PATH");
-        if (!argv[0])
-            argv[0] = QUBES_RPC_MULTIPLEXER_PATH;
-        execve(argv[0], argv, envp);
-        bool noent = errno == ENOENT;
-        PERROR("exec qubes-rpc-multiplexer");
-        _exit(noent ? QREXEC_EXIT_SERVICE_NOT_FOUND : QREXEC_EXIT_PROBLEM);
+        if (program[0] != '/') {
+            LOG(ERROR, "Program to execute not absolute path");
+            _exit(QREXEC_EXIT_PROBLEM);
+        }
+        if (i != 3 && i != 5) {
+            LOG(ERROR, "invalid number of arguments: %zu", i);
+            _exit(QREXEC_EXIT_PROBLEM);
+        }
+        const char *full_name = argv[1];
+        const char *remote_domain = argv[2];
+        const char *p = strchr(argv[1], '+');
+
+        // Change the 1 to a 0 to test the shell script generator.
+        if (!use_shell && 1) {
+            if (i == 5) {
+                EXTEND("QREXEC_REQUESTED_TARGET_TYPE=%s", argv[3]);
+                if (strcmp(argv[3], "name") == 0) {
+                    EXTEND("QREXEC_REQUESTED_TARGET=%s", argv[4]);
+                } else if (strcmp(argv[3], "keyword") == 0) {
+                    EXTEND("QREXEC_REQUESTED_TARGET_KEYWORD=%s", argv[4]);
+                } else {
+                    // requested target type unknown or not given, ignore
+                }
+            } else {
+                EXTEND_RAW("QREXEC_REQUESTED_TARGET_TYPE=");
+            }
+            EXTEND("QREXEC_SERVICE_FULL_NAME=%s", full_name);
+            EXTEND("QREXEC_REMOTE_DOMAIN=%s", remote_domain);
+            argv[1] = NULL;
+            if (p != NULL) {
+                EXTEND("QREXEC_SERVICE_ARGUMENT=%s", p + 1);
+                if (p[1] != '\0') {
+                    argv[1] = p + 1;
+                    argv[2] = NULL;
+                }
+            }
+            assert(iterator <= env_amount);
+            buf[iterator] = NULL;
+            execve(program, (char *const *)argv, buf);
+            _exit(errno == ENOENT ? QREXEC_EXIT_SERVICE_NOT_FOUND : QREXEC_EXIT_PROBLEM);
+        }
+
+        // Generate a shell command and call it with the correct arguments.
+        // Note that is_name and is_keyword are mutually exclusive.
+        bool is_name = i == 5 && strcmp(argv[3], "name") == 0;
+        bool is_keyword = i == 5 && strcmp(argv[3], "keyword") == 0;
+        const char *requested_target = i == 5 ? argv[4] : NULL;
+        argv[0] = "/bin/sh";
+        argv[1] = "-lc";
+        // Build a shell script that emulates the old multiplexer.
+        // The shell script is only built from string literals,
+        // never data from the provided command.  Reviewers are
+        // invited to check that the produced script is always
+        // valid no matter what the inputs are.  To improve
+        // readability and maintainability, separating whitespace
+        // is included in the format string, so the arguments do
+        // not need to include it.  This means that the produced
+        // script might have unnecessary whitespace, but this is
+        // harmless.
+        if (asprintf((char **)(argv + 2),
+                    // Unset the variables in case the user's shell init code did something weird,
+                    // like mark them read-only or make them an array.
+                    // Use set -e in case there is a problem.  Note that sh -el does not work
+                    // because the startup files are not compatible with -e.
+                    "set -e\n"
+                    "unset QREXEC_SERVICE_FULL_NAME QREXEC_REMOTE_DOMAIN QREXEC_REQUESTED_TARGET_TYPE "
+                    "QREXEC_SERVICE_ARGUMENT QREXEC_REQUESTED_TARGET QREXEC_REQUESTED_TARGET_KEYWORD\n"
+                    "export \"QREXEC_SERVICE_FULL_NAME=$2\" \"QREXEC_REMOTE_DOMAIN=$3\" QREXEC_REQUESTED_TARGET_TYPE=%s %s\n"
+                    // This could just be written as ${4:+"$4"}, which works just as well.
+                    // I decided to make the C code more complicated so the shell has less work
+                    // to do.
+                    "exec \"$1\" %s",
+                    is_keyword ? "keyword QREXEC_REQUESTED_TARGET_KEYWORD=\"$5\"" :
+                    (is_name ? "name QREXEC_REQUESTED_TARGET=\"$5\"" : ""),
+                    p == NULL ? "" : "\"QREXEC_SERVICE_ARGUMENT=$4\"",
+                    (p != NULL && p[1] != '\0') ? "\"$4\"" : "") < 0)
+            goto bad_asprintf;
+        argv[3] = "sh";
+        argv[4] = program;
+        argv[5] = full_name;
+        argv[6] = remote_domain;
+        argv[7] = p ? p + 1 : "";
+        argv[8] = requested_target;
+        argv[9] = NULL;
+        execve("/bin/sh", (char *const *)argv, buf);
+        /* /bin/sh should always exist */
+        _exit(QREXEC_EXIT_PROBLEM);
+bad_asprintf:
+        PERROR("asprintf");
+        _exit(QREXEC_EXIT_PROBLEM);
+    } else {
+        assert(strncmp(cmd, RPC_REQUEST_COMMAND, RPC_REQUEST_COMMAND_LEN) != 0);
     }
 }
 
@@ -94,7 +231,9 @@ void fix_fds(int fdin, int fdout, int fderr)
             fdin, fdout, fderr);
         _exit(125);
     }
-    if (dup2(fdin, 0) < 0 || dup2(fdout, 1) < 0 || dup2(fderr, 2) < 0) {
+    if ((fdin != 0 && dup2(fdin, 0) < 0) ||
+        (fdout != 1 && dup2(fdout, 1) < 0) ||
+        (fderr != 2 && dup2(fderr, 2) < 0)) {
         PERROR("dup2");
         abort();
     }
@@ -112,7 +251,8 @@ void fix_fds(int fdin, int fdout, int fderr)
         close(i);
 }
 
-static int do_fork_exec(const char *user,
+static int do_fork_exec(const char *prog,
+        const char *user,
         const char *cmdline,
         int *pid,
         int *stdin_fd,
@@ -124,8 +264,12 @@ static int do_fork_exec(const char *user,
 #define SOCK_CLOEXEC 0
 #endif
     if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, inpipe) ||
-            socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, outpipe) ||
-            (stderr_fd && socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, errpipe))) {
+            socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, outpipe)) {
+        PERROR("socketpair");
+        /* FD leaks do not matter, we exit soon anyway */
+        return -2;
+    }
+    if (stderr_fd != NULL && pipe2(errpipe, O_CLOEXEC) != 0) {
         PERROR("socketpair");
         /* FD leaks do not matter, we exit soon anyway */
         return -2;
@@ -144,7 +288,7 @@ static int do_fork_exec(const char *user,
                 fix_fds(inpipe[0], outpipe[1], 2);
 
             if (exec_func != NULL)
-                exec_func(cmdline, user);
+                exec_func(prog, cmdline, user);
             abort();
         default:
             retval = 0;
@@ -450,7 +594,7 @@ struct qrexec_parsed_command *parse_qubes_rpc_command(
             goto err;
         }
 
-        size_t const descriptor_len = (size_t)(end - start);
+        size_t const descriptor_len = cmd->service_descriptor_len = (size_t)(end - start);
         if (descriptor_len > MAX_SERVICE_NAME_LEN) {
             LOG(ERROR, "Command too long (length %zu)", descriptor_len);
             goto err;
@@ -538,7 +682,9 @@ int execute_parsed_qubes_rpc_command(
         int *stdout_fd, int *stderr_fd, struct buffer *stdin_buffer) {
     if (cmd->service_descriptor) {
         // Proper Qubes RPC call
-        int find_res = find_qrexec_service(cmd, stdin_fd, stdin_buffer);
+        char file_path[QUBES_SOCKADDR_UN_MAX_PATH_LEN];
+        struct buffer buf = { .data = file_path, .buflen = (int)sizeof(file_path) };
+        int find_res = find_qrexec_service(cmd, stdin_fd, stdin_buffer, &buf);
         if (find_res != 0) {
             assert(find_res < 0);
             return find_res;
@@ -550,12 +696,12 @@ int execute_parsed_qubes_rpc_command(
             *pid = 0;
             return 0;
         }
-        return do_fork_exec(cmd->username, cmd->command,
+        return do_fork_exec(buf.data, cmd->username, cmd->command,
                            pid, stdin_fd, stdout_fd, stderr_fd);
     } else {
         // Legacy qrexec behavior: spawn shell directly
-        return do_fork_exec(cmd->username, cmd->command,
-                           pid, stdin_fd, stdout_fd, stderr_fd);
+        return do_fork_exec(NULL, cmd->username, cmd->command,
+                            pid, stdin_fd, stdout_fd, stderr_fd);
     }
 }
 static bool validate_port(const char *port) {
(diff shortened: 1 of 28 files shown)


marmarek:
Ugh, upgrade will require a bit more care. Right now, just after upgrading dom0, qrexec calls (to dom0) fails for currently running VMs:
```
2024-06-28 13:11:55.919 qrexec-daemon[26518]: exec.c:85:exec_qubes_rpc_if_requested: exec qubes-rpc-multiplexer: No such file or directory
```

Release upgrade script will need to handle this case.

marmarek:
Maybe keep the script in repo, and remove only later (R4.4?) ?
I noticed the problem in dom0, but it probably affect domUs too.

DemiMarie:
> Maybe keep the script in repo, and remove only later (R4.4?) ? I noticed the problem in dom0, but it probably affect domUs too.

Have `systemd-tmpfiles` delete it after reboot?  Breaks downgrades, though…

Keeping it present (but unused) until R4.4 is the simplest option.

marmarek (changes requested):
I'm not sure what changed since last test, but now running services as root (or more likely - non-default user that don't use fork-server) fails due to SELinux:
```
user@dom0:~$ qvm-run -p  --service sys-net qubes.GetDate; echo $?
2024-09-27T04:31:15+00:00
0
user@dom0:~$ qvm-run -p -u root --service sys-net qubes.GetDate; echo $?
125
user@dom0:~$ qvm-run -p sys-net sudo dmesg|grep denied
[   18.800469] audit: type=1400 audit(1727411218.085:96): avc:  denied  { entrypoint } for  pid=739 comm="qrexec-agent" path="/etc/qubes-rpc/qubes.GetDate" dev="xvda3" ino=971559 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
```

DemiMarie:
> I'm not sure what changed since last test, but now running services as root (or more likely - non-default user that don't use fork-server) fails due to SELinux:
> 
> ```
> user@dom0:~$ qvm-run -p  --service sys-net qubes.GetDate; echo $?
> 2024-09-27T04:31:15+00:00
> 0
> user@dom0:~$ qvm-run -p -u root --service sys-net qubes.GetDate; echo $?
> 125
> user@dom0:~$ qvm-run -p sys-net sudo dmesg|grep denied
> [   18.800469] audit: type=1400 audit(1727411218.085:96): avc:  denied  { entrypoint } for  pid=739 comm="qrexec-agent" path="/etc/qubes-rpc/qubes.GetDate" dev="xvda3" ino=971559 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
> ```

Two options:

1. Change contexts on `/etc/qubes-rpc/*` (preferred).
2. Allow `etc_t` to be an entrypoint (but `etc_t` should not really be used for executables).

Is option 1 okay with you?  It would require a relabel when the user upgrades.

marmarek:
Yes, option 1 sounds fine. I think there is some rpm magic handling relabeling automatically already.

DemiMarie:
> Yes, option 1 sounds fine. I think there is some rpm magic handling relabeling automatically already.

That only works for TemplateVMs, not `/usr/local/etc/qubes-rpc/*` in TemplateBasedVMs.

marmarek:
> That only works for TemplateVMs, not `/usr/local/etc/qubes-rpc/*` in TemplateBasedVMs.

Right... I guess we need something like policy version flag in `/rw` to detect the need to relabel? I'd say keep a version number inside `/rw/.autorelabel`, but that's incompatible with `ConditionPathExists` (or any other Condition), so maybe rename the flag instead? `/rw/.autorelabel-done-version-1`?

marmarek:
This still fails:
```
[2025-01-04 06:39:04] [  188.534597] audit: type=1400 audit(1735972744.727:200): avc:  denied  { entrypoint } for  pid=1099 comm="qrexec-agent" path="/etc/qubes-rpc/qubes.VMShell" dev="xvda3" ino=917703 scontext=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
```
I haven't checked yet if it's update problem (I did merged #184 already, so it should work...), or wrong file context in policy in the first place.

selinux/qubes-core-qrexec.fc
```diff
@@ -1 +1,3 @@
 /usr/lib/qubes/qrexec-agent -- gen_context(system_u:object_r:qubes_qrexec_agent_exec_t,s0)
+/etc/qubes-rpc/qrexec/[^/]+ -- gen_context(system_u:object_r:bin_t,s0)
```
marmarek, line 2:
```suggestion
/etc/qubes-rpc/[^/]+ -- gen_context(system_u:object_r:bin_t,s0)
```
and similar below

marmarek:
Manual `restorecon` call doesn't help, likely due to the above wrong path

qrexec/client.py
```diff
@@ -102,11 +101,6 @@ def make_command(dest, rpcname, arg):
     if arg is not None:
         rpcname = f"{rpcname}+{arg}"
 
-    if VERSION == "dom0" and dest == "dom0":
-        # Invoke qubes-rpc-multiplexer directly. This will work for non-socket
-        # services only.
-        return [RPC_MULTIPLEXER, rpcname, "dom0"]
-
```
marmarek, line 109: Well, this needs a replacement, otherwise calling dom0->dom0 services fail. For example policy editor crashes with:
```
subprocess.CalledProcessError: Command '['/usr/bin/qrexec-client', '-d', 'dom0', 'DEFAULT:QUBESRPC policy.List dom0']' returned non-zero exit status 1.
```

marmarek:
Several of the failures suggest there is something wrong with the environment, at least PATH (see resize-rootfs failing by not finding `sfdisk`), but very likely more. Probably at least env set by PAM should be included (this should cover PATH I think), but there may be more needed (session bus address? various XDG vars etc?).

But there are also cases where I don't see what went wrong (which may mean missing error log somewhere):
```
# Test died: command 'qvm-run --no-gui -p -u root sys-net "command -v curl || (apt-get update --allow-releaseinfo-change && apt-get -y install curl ) || dnf install -y curl"' failed at qubesos/tests/suspend.pm line 30.
```
nothing on stderr from this command and all the logs from sys-net at the time:
```
[2025-01-04 07:17:14] [  294.547640] qrexec-agent[1104]: 2025-01-04 07:17:15.428 qrexec-agent[1104]: qrexec-agent-data.c:274:handle_new_process_common: executed: root:QUBESRPC qubes.VMShell dom0 (pid 1106)
[2025-01-04 07:17:14] [  294.921660] audit: type=1100 audit(1735993035.467:169): pid=1106 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 msg='op=PAM:authentication grantors=pam_rootok acct="root" exe="/usr/lib/qubes/qrexec-agent" hostname=? addr=? terminal=? res=success'
[2025-01-04 07:17:14] [  294.922201] audit: type=1103 audit(1735993035.468:170): pid=1106 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 msg='op=PAM:setcred grantors=pam_rootok acct="root" exe="/usr/lib/qubes/qrexec-agent" hostname=? addr=? terminal=? res=success'
[2025-01-04 07:17:14] [  294.922396] audit: type=1006 audit(1735993035.468:171): pid=1106 uid=0 subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 old-auid=4294967295 auid=0 tty=(none) old-ses=4294967295 ses=8 res=1
[2025-01-04 07:17:14] [  294.922535] audit: type=1300 audit(1735993035.468:171): arch=c000003e syscall=1 success=yes exit=1 a0=3 a1=7fffcefe5380 a2=1 a3=0 items=0 ppid=1104 pid=1106 auid=0 uid=0 gid=0 euid=0 suid=0 fsuid=0 egid=0 sgid=0 fsgid=0 tty=(none) ses=8 comm="qrexec-agent" exe="/usr/lib/qubes/qrexec-agent" subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 key=(null)
[2025-01-04 07:17:14] [  294.922851] audit: type=1327 audit(1735993035.468:171): proctitle="/usr/lib/qubes/qrexec-agent"
[2025-01-04 07:17:14] [  294.927077] audit: type=2300 audit(1735993035.473:172): pid=1106 uid=0 auid=0 ses=8 subj=system_u:system_r:local_login_t:s0-s0:c0.c1023 msg='op=pam_selinux default-context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 selected-context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 exe="/usr/lib/qubes/qrexec-agent" hostname=? addr=? terminal=? res=success'
[2025-01-04 07:17:15] [  294.824812] systemd[1]: Created slice user-0.slice - User Slice of UID 0.
[2025-01-04 07:17:15] [  294.841070] systemd[1]: Starting user-runtime-dir@0.service - User Runtime Directory /run/user/0...
[2025-01-04 07:17:15] [  294.903360] systemd[1]: Finished user-runtime-dir@0.service - User Runtime Directory /run/user/0.
[2025-01-04 07:17:15] [  295.239406] audit: type=1130 audit(1735993035.785:173): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=user-runtime-dir@0 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
[2025-01-04 07:17:15] [  294.909066] systemd[1]: Starting user@0.service - User Manager for UID 0...
[2025-01-04 07:17:15] [  295.289896] audit: type=1101 audit(1735993035.835:174): pid=1110 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='op=PAM:accounting grantors=pam_unix acct="root" exe="/usr/lib/systemd/systemd-executor" hostname=? addr=? terminal=? res=success'
[2025-01-04 07:17:15] [  295.289981] audit: type=1103 audit(1735993035.835:175): pid=1110 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='op=PAM:setcred grantors=? acct="root" exe="/usr/lib/systemd/systemd-executor" hostname=? addr=? terminal=? res=failed'
[2025-01-04 07:17:15] [  295.290052] audit: type=2300 audit(1735993035.836:176): pid=1110 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='op=pam_selinux default-context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 selected-context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023 exe="/usr/lib/systemd/systemd-executor" hostname=? addr=? terminal=? res=success'
[2025-01-04 07:17:15] [  295.010562] systemd-logind[530]: New session 9 of user root.
[2025-01-04 07:17:15] [  295.012654] (systemd)[1110]: pam_unix(systemd-user:session): session opened for user root(uid=0) by root(uid=0)
[2025-01-04 07:17:15] [  295.335833] systemd[1110]: Queued start job for default target default.target.
[2025-01-04 07:17:15] [  295.337263] systemd[1110]: Created slice app.slice - User Application Slice.
[2025-01-04 07:17:15] [  295.338069] systemd[1110]: Created slice session.slice - User Core Session Slice.
[2025-01-04 07:17:15] [  295.338889] systemd[1110]: grub-boot-success.timer - Mark boot as successful after the user session has run 2 minutes was skipped because of an unmet condition check (ConditionUser=!@system).
[2025-01-04 07:17:15] [  295.341287] systemd[1110]: Started systemd-tmpfiles-clean.timer - Daily Cleanup of User's Temporary Directories.
[2025-01-04 07:17:15] [  295.341625] systemd[1110]: Reached target paths.target - Paths.
[2025-01-04 07:17:15] [  295.341713] systemd[1110]: Reached target timers.target - Timers.
[2025-01-04 07:17:15] [  295.344181] systemd[1110]: Starting dbus.socket - D-Bus User Message Bus Socket...
[2025-01-04 07:17:15] [  295.344561] systemd[1110]: pipewire-pulse.socket - PipeWire PulseAudio was skipped because of an unmet condition check (ConditionUser=!root).
[2025-01-04 07:17:15] [  295.345258] systemd[1110]: Listening on pipewire.socket - PipeWire Multimedia System Sockets.
[2025-01-04 07:17:15] [  295.347522] systemd[1110]: Starting systemd-tmpfiles-setup.service - Create User Files and Directories...
[2025-01-04 07:17:15] [  295.366829] systemd[1110]: Listening on dbus.socket - D-Bus User Message Bus Socket.
[2025-01-04 07:17:15] [  295.368537] systemd[1110]: Reached target sockets.target - Sockets.
[2025-01-04 07:17:15] [  295.370438] systemd[1110]: Finished systemd-tmpfiles-setup.service - Create User Files and Directories.
[2025-01-04 07:17:15] [  295.371498] systemd[1110]: Reached target basic.target - Basic System.
[2025-01-04 07:17:15] [  295.371677] systemd[1110]: pipewire.service - PipeWire Multimedia Service was skipped because of an unmet condition check (ConditionPathExists=!/var/run/qubes-service/minimal-netvm).
[2025-01-04 07:17:15] [  295.372940] systemd[1110]: split-gpg2-client.service - split-gpg2 client was skipped because of an unmet condition check (ConditionPathExists=/run/qubes-service/split-gpg2-client).
[2025-01-04 07:17:15] [  295.373314] systemd[1110]: Reached target default.target - Main User Target.
[2025-01-04 07:17:15] [  295.374533] systemd[1110]: wireplumber.service - Multimedia Service Session Manager was skipped because of an unmet condition check (ConditionPathExists=!/var/run/qubes-service/minimal-netvm).
[2025-01-04 07:17:15] [  295.374679] systemd[1110]: Startup finished in 328ms.
[2025-01-04 07:17:15] [  295.374788] systemd[1]: Started user@0.service - User Manager for UID 0.
[2025-01-04 07:17:15] [  295.388502] systemd[1]: Started session-8.scope - Session 8 of User root.
[2025-01-04 07:17:15] [  295.393763] qrexec-agent[1106]: pam_unix(qrexec:session): session opened for user root(uid=0) by root(uid=0)
[2025-01-04 07:17:15] [  295.396441] qrexec-agent[1106]: pam_unix(qrexec:session): session closed for user root
[2025-01-04 07:17:15] [  295.398613] qrexec-agent[1104]: 2025-01-04 07:17:16.281 qrexec-agent[1104]: qrexec-agent-data.c:309:handle_new_process_common: pid 1106 exited with 125
```

marmarek:
> Right... I guess we need something like policy version flag in `/rw` to detect the need to relabel? I'd say keep a version number inside `/rw/.autorelabel`, but that's incompatible with `ConditionPathExists` (or any other Condition), so maybe rename the flag instead? `/rw/.autorelabel-done-version-1`?

And just to not forget - this is still applicable

DemiMarie:
> > Right... I guess we need something like policy version flag in `/rw` to detect the need to relabel? I'd say keep a version number inside `/rw/.autorelabel`, but that's incompatible with `ConditionPathExists` (or any other Condition), so maybe rename the flag instead? `/rw/.autorelabel-done-version-1`?
> 
> And just to not forget - this is still applicable

Better solution: run a script at each boot that checks if the SELinux policy has changed since a relabel happened and relabels the priviate volume if it has.  That is QubesOS/qubes-core-agent-linux#541 (draft because not tested yet).

marmarek:
> That is [QubesOS/qubes-core-agent-linux#541](https://github.com/QubesOS/qubes-core-agent-linux/pull/541) (draft because not tested yet).

Currently that PR checks if full rootfs relabel happened since last /rw relabel, not if policy has changed. Most policy changes will _not_ involve full relabel (rpm macros relabel only files affected by the update, using `fixfiles -C ... restore`). But yes, I agree, triggering /rw relabel if policy changed is a more generic approach. Just make sure it won't timeout VM boot on larger /rw ... (maybe using similar fixfiles trick?)

DemiMarie:
> > That is [QubesOS/qubes-core-agent-linux#541](https://github.com/QubesOS/qubes-core-agent-linux/pull/541) (draft because not tested yet).
> 
> Currently that PR checks if full rootfs relabel happened since last /rw relabel, not if policy has changed.

Indeed so, but it is easy to check if policy has changed by checking the timestamp on `/etc/selinux/$SELINUXTYPE/contexts/files/file_contexts`.

>  Most policy changes will _not_ involve full relabel (rpm macros relabel only files affected by the update, using `fixfiles -C ... restore`). But yes, I agree, triggering /rw relabel if policy changed is a more generic approach. Just make sure it won't timeout VM boot on larger /rw ... (maybe using similar fixfiles trick?)

`fixfiles` will try to relabel the whole root filesystem, which is bad, and it also excludes `/home` by default, so non-Qubes systems are actually broken!  It is also a shell script that badly needs a rewrite in a better language and almost certainly relabels far more than it needs to.

What about using the `-regex` option to `find`?

marmarek:
> `fixfiles` will try to relabel the whole root filesystem, which is bad

That's why I said "similar", not "the same".

> What about using the `-regex` option to `find`?

Maybe, but remember there are a bunch of corner cases, like `file_contexts.subs`. If going with "similar to fixfiles" approach, maybe you can at least take the policy diff part from there.

DemiMarie:
Some local testing:

- [x] dom0 ⇒ VM calls that use the fork server work fine.
- [ ] `qvm-run --user=root --service` runs the service with `PATH=/usr/local/bin:/usr/bin` (wrong).
- [ ] Running a service in `/usr/local/etc/qubes-rpc` as root does not work at all (wrong).

DemiMarie:
dom0 ⇒ dom0 calls can be solved by either continuing to use `qubes-rpc-multiplexer` or just reimplementing the functionality in pure Python.  The rest of the problems are either due to SELinux (which I will fix) or not using a login shell to spawn the process.  The latter can be worked around in two ways:

1. Always use a fork server, and create it lazily, on-demand.
2. Run the command via `/bin/sh -eulc 'exec "$@"' sh PROG ARGUMENT`.

marmarek:
> dom0 ⇒ dom0 calls can be solved by either continuing to use `qubes-rpc-multiplexer` or just reimplementing the functionality in pure Python

Both sounds good. There is one more place where `qubes-rpc-multiplexer` is used: https://github.com/QubesOS/qubes-core-admin/blob/main/qubes/vm/adminvm.py#L312-L320
TBH I'm surprised it isn't used in core-admin-client, but this one seems to re-implement it in a very limited way (see `core-admin-client/qubesadmin/app.py:QubesLocal.run_service()`).
I don't see any usages outside of python, so doing that in that python function directly should be fine too. 

> Always use a fork server, and create it lazily, on-demand.

How would it be started? The whole point of the fork-server is starting processes inside user session that is started independently of qrexec (in most cases by gui-agent, but in some cases by full lightdm or similar). Starting it from qrexec doesn't solve the issue of not having full user session...
The current code already opens user session for isolated calls (see `agent/qrexec-agent.c:do_exec()`) and as seen this is not enough, starting fork-server the same way wouldn't help - it would need to be adjusted anyway, at which point fork-server doesn't help with this issue (starting the service directly after the adjustments should also work).
On the other hand, fork-server may avoid having user session started over and over for repeated calls (not really common for non-default user). But also, IMO if going that way, it should have a timeout to not keep several processes running indefinitely if a service was called once. Sounds like a separate feature.

> Run the command via `/bin/sh -eulc 'exec "$@"' sh PROG ARGUMENT`.

Will that solve setting up environment? Including /etc/environment.d This sounds like a simpler approach. It can be improved later if necessary.

See also https://github.com/QubesOS/qubes-issues/issues/3434 - kinda related.

libqrexec/exec.c
```diff
@@ -47,44 +48,127 @@ void register_exec_func(do_exec_t *func) {
...
+    if (program[0] != '/') {
+        LOG(ERROR, "Program to execute not absolute path");
+        _exit(QREXEC_EXIT_PROBLEM);
+    }
+    if (use_shell) {
+        argv[3] = argv[0];
+        argv[4] = argv[1];
+        argv[5] = NULL;
+        argv[0] = "-sh";
+        argv[1] = "-lcunset LESSOPEN LESSCLOSE&&exec \"$@\"";
```
marmarek, line 162: This is definitely _not_ the place for some random harcoded config overrides.

marmarek:
Besides requested change above, what is the state of the three issues you identified at https://github.com/QubesOS/qubes-core-qrexec/pull/141#issuecomment-2576758590

marmarek:
E

> * system_tests_gui_interactive
>   
>   * screenlocker_leaks: [Failed](https://openqa.qubes-os.org/tests/125876#step/screenlocker_leaks/5) (test died)
>     `# Test died: command 'qvm-run -p -u root work "dnf -y install mate-...`
> 
> *

All tests failed with message like this:
```
-sh: - : invalid option
```
and also black complains

qrexec/client.py
```diff
@@ -99,14 +99,11 @@ async def call_async(dest, rpcname, arg=None, *, input=None):
 
 def make_command(dest, rpcname, arg):
     assert "+" not in rpcname
+    assert " " not in rpcname
+    assert " " not in arg
```
marmarek, line 103: This assert breaks stuff:
```
  File "/usr/lib/python3.13/site-packages/qubes_config/policy_editor/policy_editor.py", line 179, in do_activate
    self.perform_setup()
    ~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/qubes_config/policy_editor/policy_editor.py", line 201, in perform_setup
    self.file_select_handler = OpenDialogHandler(
                               ~~~~~~~~~~~~~~~~~^
        self.builder, self.policy_client, self.open_policy_file
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.13/site-packages/qubes_config/policy_editor/policy_editor.py", line 83, in __init__
    for file in self.policy_client.policy_list():
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/qubes_config/policy_editor/policy_editor.py", line 134, in policy_list
    file_list = self.policy_client.policy_list()
  File "/usr/lib/python3.13/site-packages/qrexec/policy/admin_client.py", line 43, in policy_list
    return self.call("policy.List").rstrip("\n").split("\n")
           ~~~~~~~~~^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qrexec/policy/admin_client.py", line 74, in call
    return call("dom0", service_name, arg, input=payload)
  File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 52, in call
    command = make_command(dest, rpcname, arg)
  File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 103, in make_command
    assert " " not in arg
           ^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```

and in the policy daemon itself too:
```
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]: Unhandled exception in client_connected_cb
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]: transport: <_SelectorSocketTransport closed fd=10>
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]: Traceback (most recent call last):
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:   File "/usr/lib/python3.13/site-packages/qrexec/tools/qrexec_policy_daemon.py", line 125, in handle_client_connection
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     result = await handle_request(
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:              ^^^^^^^^^^^^^^^^^^^^^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:         **args, log=log, policy_cache=policy_cache
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     )
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     ^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:   File "/usr/lib/python3.13/site-packages/qrexec/tools/qrexec_policy_exec.py", line 395, in handle_request
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     return await resolution.execute()
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:   File "/usr/lib/python3.13/site-packages/qrexec/tools/qrexec_policy_exec.py", line 105, in execute
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     ask_response = await call_socket_service(
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:         guivm, service, source_domain, params
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     )
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     ^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:   File "/usr/lib/python3.13/site-packages/qrexec/server.py", line 119, in call_socket_service_remote
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     output_data = await call_async(remote_domain, service, input=input_data)
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:   File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 74, in call_async
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     command = make_command(dest, rpcname, arg)
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:   File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 103, in make_command
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:     assert " " not in arg
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]:            ^^^^^^^^^^^^^^
Jan 19 01:39:50.538866 dom0 qrexec-policy-daemon[1691]: TypeError: argument of type 'NoneType' is not iterable
```

marmarek:
> * TC_00_USBProxy_debian-12-xfce: [test_000_attach_detach](https://openqa.qubes-os.org/tests/126009#step/TC_00_USBProxy_debian-12-xfce/1) (failure)
>   `AssertionError: 1 != 0 : qubes.USBAttach call failed`
> 
> *

All USB proxy tests failed, but I'm not sure why. Besides qubes.USBAttach reporting failure, I do see qubes.USB call frontend->backend executed, and at least backend seems to do something:
```
[2025-01-19 03:01:51] [   14.253882] usbip-host 1-1: usbip-host: register new device (bus 1 dev 2)
[2025-01-19 03:01:51] [   14.258881] usbip-host 1-1: stub up
[2025-01-19 03:01:51] [   14.261999] usbip-host 1-1: recv a header, 0
[2025-01-19 03:01:51] [   14.376495] usbip-host 1-1: reset high-speed USB device number 2 using dummy_hcd
[2025-01-19 03:01:52] [   14.509223] usbip-host 1-1: device reset
```
it only fails later, on cleanup (still on backend side):
```
[2025-01-19 03:01:52] [   15.064752] qubes.USB+1-1-test-inst-frontend[1009]: Failed to remove qubesdb-write
[2025-01-19 03:01:52] [   15.065316] qubes.USB+1-1-test-inst-frontend[1009]: Failed to remove
```
on frontend I see vhci controller being initialized, but I don't see actual device appearing.

DemiMarie:
> > * TC_00_USBProxy_debian-12-xfce: [test_000_attach_detach](https://openqa.qubes-os.org/tests/126009#step/TC_00_USBProxy_debian-12-xfce/1) (failure)
> >   `AssertionError: 1 != 0 : qubes.USBAttach call failed`
> 
> All USB proxy tests failed, but I'm not sure why. Besides qubes.USBAttach reporting failure, I do see qubes.USB call frontend->backend executed, and at least backend seems to do something:

Cause is that `qubes.USBAttach` runs `exec 2>"/proc/$SERVICE_ATTACH_PID/fd/2"`, which fails because stderr is now an `AF_UNIX` socket.  The fix is for `fix_fds` to use a pipe for stderr instead of a socket.

DemiMarie:
I tested this commit and USB device attach works.

marmarek:
Policy prompt in sys-gui doesn't work. This is what I find in the log there:
```
2025-01-22 14:12:22.454 qrexec-fork-server[5441]: qrexec-agent-data.c:274:handle_new_process_common: executed: QUBESRPC policy.Ask+ dom0 (pid 0)
2025-01-22 14:12:22.455 qrexec-fork-server[5441]: qrexec-agent-data.c:311:handle_new_process_common: Socket service closed
Unhandled exception in client_connected_cb
transport: <gbulb.transports.SocketTransport object at 0x7c8c599de330>
  File "/usr/lib/python3.13/site-packages/qrexec/server.py", line 71, in _client_connected
    response = await self.handle_request(params, service, source_domain)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qrexec/tools/qrexec_policy_agent.py", line 563, in handle_request
    raise ValueError("unknown service name: {}".format(service))
ValueError: unknown service name: policy.Ask+
2025-01-22 14:12:22.463 qrexec-fork-server[5445]: qrexec-agent-data.c:274:handle_new_process_common: executed: QUBESRPC policy.Notify+ dom0 (pid 0)
2025-01-22 14:12:22.464 qrexec-fork-server[5445]: qrexec-agent-data.c:311:handle_new_process_common: Socket service closed
Unhandled exception in client_connected_cb
transport: <gbulb.transports.SocketTransport object at 0x7c8c599de5d0>
  File "/usr/lib/python3.13/site-packages/qrexec/server.py", line 71, in _client_connected
    response = await self.handle_request(params, service, source_domain)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qrexec/tools/qrexec_policy_agent.py", line 563, in handle_request
    raise ValueError("unknown service name: {}".format(service))
ValueError: unknown service name: policy.Notify+
```
The issue does not apply to dom0.

There are two issues here:
1. The policy agent not dealing with the +
2. The behavior change that it got appended.

The first one is just a bug to fix.
The second one is likely intended change. But it needs to be clearly documented. There needs to be info of this change in the commit message that introduces the change, but also relevant qubes-doc page needs an info that starting with R4.3 "+" is always added there (BTW are exec-based services affected too? like value of some of the env variables set by the qrexec). And also wants mentioning in the release notes. For the last one, the best thing might be opening a separate issue for this change, adding "release notes" label, and closing it with the commit that does the change.

marmarek:
As for the performance tests, the most visible difference is for the case of the fork-server - there the difference is over 2x. This is the most common case, so I'm happy :) But some other cases got improved too. The no-fork-server case (calls using root) got slightly worse in some cases - not exactly sure why, but also the difference is quite small.
Base results (without this PR): https://openqa.qubes-os.org/tests/126236/file/system_tests-tests-qubes.tests.integ.qrexec_perf.log
Results with this PR: https://openqa.qubes-os.org/tests/126528/file/system_tests-tests-qubes.tests.integ.qrexec_perf.log

DemiMarie:
This should be ready to merge @marmarek, assuming that no tests fail unexpectedly in openQA.  The PR description is outdated, though.  If possible it would be better to just do a fast-forward merge.

marmarek:
> Policy prompt in sys-gui doesn't work. This is what I find in the log there:

Exact same error is still the with the most recent version of this PR

DemiMarie:
I will add some tests to first figure out what the current behavior is so that I can ensure it is preserved.

DemiMarie:
> > Policy prompt in sys-gui doesn't work. This is what I find in the log there:
> 
> Exact same error is still there with the most recent version of this PR

I fixed the bug in #187 but `git rebase` added it back in the form of an unwanted commit.  I fixed this by removing the unwanted commit.

marmarek:
Now indeed sys-gui tests are green :)

marmarek:
> All variables with names beginning with QREXEC are stripped from the
environment, except for QREXEC_SERVICE_PATH.  This is a change in
behavior compared to the current code.

Why actually doing it? Is it intended to remove some specific variable(s)?

DemiMarie:
> > All variables with names beginning with `QREXEC` are stripped from the
> > environment, except for `QREXEC_SERVICE_PATH`.  This is a change in
> > behavior compared to the current code.
> 
> Why actually doing it? Is it intended to remove some specific variable(s)?

The intent is to ensure that with the exception of `QREXEC_SERVICE_PATH`, all `QREXEC*` variables are unset if qrexec did not set them explicitly, even if the user has them set in the environment somewhere.  This is especially useful for testing, as it ensures that `QREXEC*` variables from the user’s environment don’t leak into the test environment.

marmarek:
The way you described it may suggest various `QREXEC_*` variables like `QREXEC_REMOTE_DOMAIN` stopped being supported, which would be bad.

For testing, you don't need to modify release behavior to avoid those variables - simply exclude them in tests when calling various qrexec tools.

Anyway, I'm not strongly against this change, as it feels harmless (if not useful), but it needs better description. But also, would it be feasible to add some logging when excluding a variable that isn't going to be set a moment later (so, a variable that would be "leaked" to the service)? That would help finding unintended effects of this change.

libqrexec/exec.c
```diff
@@ -47,44 +48,170 @@ void register_exec_func(do_exec_t *func) {
...
+    }
+    if (i != 3 && i != 5) {
+        LOG(ERROR, "invalid number of arguments: %zu", i);
+        _exit(QREXEC_EXIT_PROBLEM);
+    }
+    const char *full_name = argv[1];
+    const char *remote_domain = argv[2];
+    const char *p = strchr(argv[1], '+');
+
+    // Change the 0 to a 1 to test the shell script generator.
```
marmarek, line 137: This comment doesn't match the next line

marmarek (approved):

PR #142, opened 2024-04-11
Support socket services with `MSG_JUST_EXEC`

DemiMarie:
This separates _finding_ a qrexec service (and, if necessary, connecting to a socket) from actually _executing_ the service and processing I/O. This allows qrexec-agent to handle `MSG_JUST_EXEC` correctly, because it can connect to a socket-based service without having to be prepared for an executable service to be run immediately.  Instead, qrexec-agent spawns an executable service using its own spawning routines that do support `MSG_JUST_EXEC`.

This also adds tests for RPC services (both executable and socket) with `MSG_JUST_EXEC`.  Previously, these were not tested, despite being essential to a running system.  For instance, the Qubes app menu uses `MSG_JUST_EXEC` to invoke the qubes.StartApp service every time it launches an application in a non-disposable VM from dom0.

### DIFF
--- a/agent/qrexec-agent-data.c
+++ b/agent/qrexec-agent-data.c
@@ -136,6 +136,16 @@ static int handle_just_exec(struct qrexec_parsed_command *cmd)
 
     if (cmd == NULL)
         return 127;
+
+    if (cmd->service_descriptor) {
+        int socket_fd;
+        struct buffer stdin_buffer;
+        buffer_init(&stdin_buffer);
+        if (!find_qrexec_service(cmd, &socket_fd, &stdin_buffer))
+            return 127;
+        if (socket_fd != -1)
+            return write_all(socket_fd, stdin_buffer.data, stdin_buffer.buflen) ? 0 : 127;
+    }
     switch (pid = fork()) {
         case -1:
             PERROR("fork");
--- a/qrexec/tests/socket/agent.py
+++ b/qrexec/tests/socket/agent.py
@@ -27,6 +27,7 @@ import struct
 import getpass
 import itertools
 import asyncio
+import shlex
 
 import psutil
 import pytest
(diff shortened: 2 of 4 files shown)


marmarek (changes requested):
This has now conflicts on tests (some of this changes were also in #144 )

agent/qrexec-agent-data.c
```diff
@@ -136,6 +136,20 @@ static int handle_just_exec(struct qrexec_parsed_command *cmd)
...
+    struct buffer stdin_buffer;
+    buffer_init(&stdin_buffer);
+    if (cmd->service_descriptor) {
+        int socket_fd;
+        if (!find_qrexec_service(cmd, &socket_fd, &stdin_buffer, &buf))
+            return 127;
+        if (socket_fd != -1)
+            return write_all(socket_fd, stdin_buffer.data, stdin_buffer.buflen) ? 0 : 127;
+    } else {
+        buf.data = NULL;
```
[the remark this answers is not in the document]
DemiMarie: This is based on code written for #141.  Fixed.

qrexec/tests/socket/agent.py
```diff
@@ -148,28 +151,52 @@ def test_handshake(self):
...
-        cmd = ("touch " + os.path.join(self.tempdir, "new_file")).encode(
-            "ascii"
-        )
         dom0.send_message(
             qrexec.MSG_JUST_EXEC,
             struct.pack("<LL", self.target_domain, self.target_port)
             + user
             + b":"
             + cmd
-            + b"\0",
```
marmarek, line 168: Why removing it here and forcing every caller to include \0 in the string? Are you expecting some (future?) call to not have this final NUL byte?

DemiMarie: I think it made some assertions easier, but reverted.

marmarek (approved):

PR #143, opened 2024-04-12
Avoid using alarm(2) for timeouts

DemiMarie:
libvchan has much better ways of dealing with timeouts now, so use them instead of calling an async signal unsafe function from a signal handler.  Both qrexec-agent-data.c and qrexec-daemon-common.c reimplemented timeouts, so write common code in libqrexec for both to use.

Based on #136 which is already approved.

### DIFF
--- a/daemon/qrexec-client.c
+++ b/daemon/qrexec-client.c
@@ -143,57 +143,6 @@ bad_c_param:
     exit(1);
 }
 
-static const long BILLION_NANOSECONDS = 1000000000L;
-
-static void wait_for_vchan_client_with_timeout(libvchan_t *conn, time_t timeout) {
-    struct timespec end_tp, now_tp, timeout_tp;
-
-    if (timeout && clock_gettime(CLOCK_MONOTONIC, &end_tp)) {
-        PERROR("clock_gettime");
-        exit(1);
-    }
-    assert(end_tp.tv_nsec >= 0 && end_tp.tv_nsec < BILLION_NANOSECONDS);
-    end_tp.tv_sec += timeout;
-    int const fd = libvchan_fd_for_select(conn);
-    while (conn && libvchan_is_open(conn) == VCHAN_WAITING) {
-        if (timeout) {
-            bool did_timeout = true;
-            struct pollfd fds = { .fd = fd, .events = POLLIN | POLLHUP, .revents = 0 };
-
-            /* calculate how much time left until connection timeout expire */
-            if (clock_gettime(CLOCK_MONOTONIC, &now_tp)) {
-                PERROR("clock_gettime");
-                exit(1);
-            }
-            assert(now_tp.tv_nsec >= 0 && now_tp.tv_nsec < BILLION_NANOSECONDS);
-            if (now_tp.tv_sec <= end_tp.tv_sec) {
-                timeout_tp.tv_sec = end_tp.tv_sec - now_tp.tv_sec;
-                timeout_tp.tv_nsec = end_tp.tv_nsec - now_tp.tv_nsec;
-                if (timeout_tp.tv_nsec < 0) {
-                    timeout_tp.tv_nsec += BILLION_NANOSECONDS;
-                    timeout_tp.tv_sec--;
-                }
-                did_timeout = timeout_tp.tv_sec < 0;
-            }
-            switch (did_timeout ? 0 : ppoll(&fds, 1, &timeout_tp, NULL)) {
-            case -1:
-                if (errno == EINTR)
-                    break;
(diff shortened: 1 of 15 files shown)


marmarek:
This is a welcome change, but requires careful testing. I think the easiest way to test is to pause relevant VM before confirming qrexec policy prompt. Pausing either source or destination should allow testing connection timeout handling on either vchan server or client side. Similarly pause target qube before making a call from dom0 to test that part too. And finally, to test qrexec-daemon startup timeout, you can pause qube during startup, or simply "hang" it there by adding `init=/bin/sh` (or similar) to its kernelopts.

DemiMarie:
> This is a welcome change, but requires careful testing. I think the easiest way to test is to pause relevant VM before confirming qrexec policy prompt. Pausing either source or destination should allow testing connection timeout handling on either vchan server or client side. Similarly pause target qube before making a call from dom0 to test that part too. And finally, to test qrexec-daemon startup timeout, you can pause qube during startup, or simply "hang" it there by adding `init=/bin/sh` (or similar) to its kernelopts.

Tested so far:

- [x] VM ⇒ dom0 calls
  - [x] Pause VM before confirmation prompt
  - [x] Resume VM after dom0 times out
- [x] VM ⇒ VM calls
  - [x] Call to paused VM
  - [x] Call by paused VM

I believe this leaves only the case of dom0 being slow, but I don’t think `qrexec-agent` differentiates between calls to dom0 and calls to a VM.

marmarek:
Main point of the alarm() call you replace is handling VM startup timeout, so that needs testing too.

DemiMarie:
> `init=/bin/sh` (or similar) to its kernelopts.

doesn’t work, VM starts normally

marmarek:
then maybe some systemd-related option, to avoid starting qrexec service for example?

DemiMarie:
> then maybe some systemd-related option, to avoid starting qrexec service for example?

`rdinit=/bin/bash` works.  Turns out that `init=` is ignored by Linux if one is using an initramfs (which Qubes OS is).

I also pushed a bunch of related changes to signal handling and timeouts.  `qrexec-daemon` now handles VM startup timeouts by checking the return value `libvchan_client_init_async` and notifying the parent process via a pipe.  A different value is written to the pipe to indicate success.

DemiMarie:
CI failure is a due to a flaky test.  #144 should fix that.

DemiMarie:
PipelineRetryFailed

daemon/qrexec-client.c
```diff
@@ -365,9 +314,16 @@ int main(int argc, char **argv)
                     VCHAN_BUFFER_SIZE, VCHAN_BUFFER_SIZE);
             if (!data_vchan) {
                 LOG(ERROR, "Failed to start data vchan server");
-                exit(1);
+                rc = 126;
```
marmarek, line 317: Please don't abuse 126 code for generic errors. If no better idea, you can leave it at 1 for now (here and in all the other places below). But, besides something from `sysexits.h` idea from the other PR, maybe declare 125 as "service call error"? Should be unlikely to conflict with the actual service exit code, I hope.

marmarek (changes requested):

daemon/qrexec-client.c
marmarek: That's an interesting proposal, but due to exactly what that would mean, it can easily confuse various applications that (possibly also some shells...) that it was indeed killed by a signal, or if something takes just 7 bits (excluding the "killed" case) it will even take it as a success. Yes, such applications are buggy. But also, lets not try running into such bugs.

DemiMarie: That is fair.

I think using `sysexits.h` codes is the best thing to do, except maybe if systemd defines a better standard.  That said, it does leak a small amount of information to the calling program, most notably if there was an out-of-memory condition.  Is that a problem?

marmarek: > codes

A singular code. A single one for any type of failing to start the allowed existing service call.

libqrexec/qrexec.h
```diff
@@ -173,4 +173,11 @@ enum {
 // support only very small configuration files,
 #define MAX_CONFIG_SIZE 4096
 
+// Exit codes
+
+// Service call refused
+#define QREXEC_EXIT_REQUEST_REFUSED 126
+// Problem with qrexec itself
+#define QREXEC_EXIT_PROBLEM 128
```
marmarek, line 181: Please use some code in 1-125 range (126 and 127 are already used)

(putting as a separate comment for visibility)

marmarek (changes requested):

marmarek (approved):

PR #146, opened 2024-04-21
Reject invalid service targets

DemiMarie:
If a service target is syntactically incorrect (and would be mangled by qrexec-daemon), reject it outright instead.  This also adds support for re-validating data from the policy daemon and the command line, to ensure that produced commands are correct by construction.  However, this is turned off in the build, because such changes have a nonzero risk.

Fixes: QubesOS/qubes-issues#9140

### DIFF
--- a/daemon/qrexec-daemon.c
+++ b/daemon/qrexec-daemon.c
@@ -791,26 +791,6 @@ static int find_policy_pending_slot(void) {
     return -1;
 }
 
-static void sanitize_name(char * untrusted_s_signed, char *extra_allowed_chars)
-{
-    unsigned char * untrusted_s;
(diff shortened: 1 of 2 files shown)


DemiMarie:
> [Not a good idea to change it now](https://github.com/QubesOS/qubes-issues/issues/9140#issuecomment-2068217557), we can reconsider for R5.0.
> 
> Closing for now.

I’ll create a new PR with changes to tests, so that the current behavior is at least tested for.

marmarek:
Reopened as I misread the purpose. But it still isn't clear if this is applicable as R4.x update.

DemiMarie:
Closing as per https://github.com/QubesOS/qubes-issues/issues/9140#issuecomment-2068284915

PR #147, opened 2024-04-23
Avoid writing to an uninitialized file descriptor

DemiMarie:
Found by clang-tidy.  No tests, but the fix is trivial.

Fixes: 3d658cc82739 ("Use a pipe instead of signals to notify readiness")

### DIFF
--- a/daemon/qrexec-daemon.c
+++ b/daemon/qrexec-daemon.c
@@ -110,7 +110,6 @@ static const char default_user_keyword[] = "DEFAULT:";
 #define default_user_keyword_len_without_colon (sizeof(default_user_keyword)-2)
 
 static int opt_quiet = 0;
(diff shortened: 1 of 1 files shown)


marmarek:
Ouch, I noticed it only now, a9aa1c5344c41cd5684d41b38d05815af735dddd changed behavior in unintended way. It says:
> This avoids leaving the child process running.

But in case of `QREXEC_STARTUP_NOWAIT` set, it was actually intentional. `QREXEC_STARTUP_NOWAIT` in practice means no timeout, but after the mentioned commit, default timeout will be enforced in that case.
Can you add a fix to this PR (since it touches kinda related code)?

DemiMarie:
Sure.  I’ll also add a test.

marmarek (approved):

PR #137, opened 2024-03-20
Miscellaneous cleanups and bug fixes

DemiMarie:
These are all minor, backportable changes.  All should be self-contained, but cherry-picking individual patches might result in artificial conflicts.

Fixes: QubesOS/qubes-issues#9161.

PR #150, opened 2024-04-27
If skip-service-descriptor=true, do not use fork server

DemiMarie:
The fork server doesn't load the configuration, so it would send the service descriptor anyway.

Also change SELinux policy so that qrexec-agent can actually connect to TCP sockets.

Fixes: QubesOS/qubes-issues#9161


DemiMarie:
Made part of #137 to avoid merge conflicts in tests.

PR #151, opened 2024-04-28
Do not make a file description blocking while it is still in use

DemiMarie:
use_stdio_socket means that stdin and stdout FDs use the same file description, which is certainly the case for socket-based services. Before this change, it was set to false in this case.  This would cause the file description to be made blocking while it is still in use, potentially causing deadlocks.

git blame points to commit 0802df70eefcf5f897b3782313d01bc1bbc3712b ("Factor out process_child_io"), but I suspect the actual bug dates back as far as the introduction of socket-based services in 76697e30f6e19bb003d669e5bdcb4dae739ac878 ("Working socket-based qrexec").

Fixes: QubesOS/qubes-issues#9169

### DIFF
--- a/daemon/qrexec-client.c
+++ b/daemon/qrexec-client.c
@@ -72,11 +72,16 @@ static _Noreturn void do_exec(const char *prog, const char *username __attribute
     PERROR("exec bash");
     exit(1);
 }
+enum {
+    opt_socket_dir = 'd'+128,
+    opt_use_stdin_socket = 'u'+128,
+};
 
 static struct option longopts[] = {
     { "help", no_argument, 0, 'h' },
-    { "socket-dir", required_argument, 0, 'd'+128 },
+    { "socket-dir", required_argument, 0, opt_socket_dir },
     { "no-exit-code", no_argument, 0, 'E' },
+    { "use-stdin-socket", no_argument, 0, opt_use_stdin_socket },
     { NULL, 0, 0, 0 },
 };
 
@@ -96,7 +101,8 @@ _Noreturn static void usage(const char *const name)
             "  -c - connect to existing process (response to trigger service call)\n"
             "  -w timeout - override default connection timeout of 5s (set 0 for no timeout)\n"
             "  -k - kill the domain right before exiting\n"
-            "  --socket-dir=PATH -  directory for qrexec socket, default: %s\n",
+            "  --socket-dir=PATH -  directory for qrexec socket, default: %s\n"
+            "  --use-stdin-socket - use fd 0 (which must be socket) for both stdin and stdout\n",
             name ? name : "qrexec-client", QREXEC_DAEMON_SOCKET_DIR);
     exit(1);
 }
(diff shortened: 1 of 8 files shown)


DemiMarie:
@marmarek can you test this?

marmarek:
It doesn't help.

libqrexec/process_io.c
```diff
@@ -40,45 +40,40 @@ static _Noreturn void handle_vchan_error(const char *op)
...
-        if (errno == ENOTSOCK)
-            close(fd);
-        else
+    else {
+        if (shutdown(fd, SHUT_WR) == -1)
             PERROR("shutdown close_stdin");
     }
+
+    if (close_fd)
+        close(fd);
```
marmarek, line 60: Now, with restore_block=true and close_fd=false this function won't close (using any method) the FD at all. I haven't checked if such combination is possible, but this behavior feels wrong: if it can happen, then it needs to close the VM using some method. If not, documentation above wants an update (and maybe also include an assert about this?). And then the parameter name is confusing, as `restore_block` now also controls using `shutdown` or not.

marmarek: And also, previous version handled ENOTSOCK. Are you absolutely sure it can't happen? `process_io` may get the FD from the user process, so I don't think you can assume it's always a socket (even if this path would be reachable only after service sending SIGUSR1).

daemon/qrexec-client.c
```diff
@@ -72,11 +72,16 @@ static _Noreturn void do_exec(const char *prog, const char *username __attribute
...
+    opt_socket_dir = 'd'+128,
+    opt_use_stdin_socket = 'u'+128,
+};
 
 static struct option longopts[] = {
     { "help", no_argument, 0, 'h' },
-    { "socket-dir", required_argument, 0, 'd'+128 },
+    { "socket-dir", required_argument, 0, opt_socket_dir },
     { "no-exit-code", no_argument, 0, 'E' },
+    { "use-fd0-socket", no_argument, 0, opt_use_stdin_socket },
```
marmarek, line 84: Maybe `--use-stdin-socket` is a better name after all?

marmarek (changes requested):

libqrexec/process_io.c
DemiMarie: > Now, with restore_block=true and close_fd=false this function won't close (using any method) the FD at all. I haven't checked if such combination is possible, but this behavior feels wrong: if it can happen, then it needs to close the VM using some method. If not, documentation above wants an update (and maybe also include an assert about this?).

This _will_ happen for socket-based services, but the behavior is correct: the call to `close_stdin()` is effectively a no-op, and the call to `close_stdout()` closes the socket.

<https://github.com/QubesOS/qubes-core-qrexec/blob/d1f525204c5004c23fc7557433df33b3edb203cd/libqrexec/process_io.c#L307-L308>

However, I completely rewrote this in the most recent force-push.

> And then the parameter name is confusing, as `restore_block` now also controls using `shutdown` or not.

Fixed by getting rid of the parameter.

> And also, previous version handled ENOTSOCK. Are you absolutely sure it can't happen? `process_io` may get the FD from the user process, so I don't think you can assume it's always a socket (even if this path would be reachable only after service sending SIGUSR1).

Fixed.

libqrexec/process_io.c
```diff
@@ -133,10 +111,21 @@ int process_io(const struct process_io_request *req) {
     set_nonblock(stdin_fd);
     if (stdout_fd != stdin_fd)
         set_nonblock(stdout_fd);
-    else if ((stdout_fd = fcntl(stdin_fd, F_DUPFD_CLOEXEC, 3)) < 0)
-        abort(); // not worth handling running out of file descriptors
-    if (stderr_fd >= 0)
+    if (stderr_fd >= 0) {
+        assert(is_service); // if this is a client, stdin_fd is *always* -1
```
marmarek, line 115:
```suggestion
        assert(is_service); // if this is a client, stderr_fd is *always* -1
```

DemiMarie:
To avoid breaking on musl libc and to ensure correct behavior if `close_range()` isn’t supported, I decided to use a loop over `/proc/self/fd` as the fallback.  It was not hard to write.

marmarek (approved):

PR #154, opened 2024-05-01
Explain why no use after free

DemiMarie:
When looking at the code I initially thought that commit
b8f0031285d87cbef9aad9902ef1fea9126ad960 introduced a use after free
vulnerability.  It does not, but this was not clear from reading the
code, so add comments to explain why the buffer (unlike stdin_buf) can
just be uninitialized memory and will always be valid throughout
process_io() until it is freed.

Also avoid allocating two buffers when one will do.

### DIFF
--- a/libqrexec/remote.h
+++ b/libqrexec/remote.h
@@ -40,6 +40,17 @@
  * Options:
  *   replace_chars_stdout, replace_chars_stderr - remove non-printable
  *     characters from stdout/stderr
+ *
+ * stdin_buf is the buffer holding data to be written to stdin, and may
+ * be modified or reallocated.  Its content must be valid data, and on
+ * return it will be a valid buffer pointing to valid (generally different)
+ * data.
+ *
+ * buffer is scratch space _only_.  Its size must be the maximum data
+ * chunk size.  Its contents (the data pointed to) does _not_ need to be
+ * initialized, and will _not_ be anything meaningful on return.  The
+ * buffer pointer and length will not be changed and will remain valid,
+ * though.  In Rust terms: this is an &mut [MaybeUninit<u8>].
  */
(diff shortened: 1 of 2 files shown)


libqrexec/remote.h
```diff
@@ -0,0 +1,79 @@
...
+ * Options:
+ *   replace_chars_stdout, replace_chars_stderr - remove non-printable
+ *     characters from stdout/stderr
+ *
+ * stdin_buf is the buffer holding data to be written to stdin, and may
+ * be modified or reallocated.  Its content must be valid data, and on
+ * return it will be a valid buffer pointing to valid (generally different)
+ * data.
+ *
+ * buffer is scratch space _only_, and will _not_ be modified or reallocated.
```
marmarek, line 49: "will _not_ be modified" suggests the content won't be modified, which definitely is not true.

libqrexec/remote.h
```diff
@@ -0,0 +1,79 @@
...
+ * Handle data from the specified FD (cannot be -1) and send it over vchan
+ * with a given message type (MSG_DATA_STDIN/STDOUT/STDERR).
+ *
+ * Return codes:
+ *   REMOTE_ERROR - vchan error occured
+ *   REMOTE_EOF - EOF received, do not access this FD again
+ *   REMOTE_OK - some data processed, call it again when buffer space and
+ *     more data availabla
+ *
+ * buffer is scratch space _only_, and will _not_ be modified or reallocated.
```
marmarek, line 70: Same here

marmarek (changes requested):

DemiMarie:
Updated comments to explain that the _pointer_ and _length_ will remain valid, but the _data_ will not.

marmarek (approved):

PR #157, opened 2024-05-01
Fix SIGUSR1 after stdin_fd closed

DemiMarie:
Just shut it down for writing, and save it so that it can later be used for reading.

This also ensures deterministic behavior when --use-stdin-socket is used.  Previously, the socket would be shut down for reading or writing but not both, depending on whether reading or writing stopped first. This could be quite confusing for callers.  Instead, always shut down the socket for both reading and writing, guaranteeing deterministic behavior.

Fixes: a3bbcb5f4314 ("Prefer close() to shutdown()")
Fixes: QubesOS/qubes-issues#9183

### DIFF
--- a/libqrexec/process_io.c
+++ b/libqrexec/process_io.c
@@ -42,24 +42,31 @@ static _Noreturn void handle_vchan_error(const char *op)
 /*
  * Closing the file descriptors:
  *
- * - If this is a single socket used for both stdin and stdout, call shutdown()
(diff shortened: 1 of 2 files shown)


marmarek:
this fails to build

DemiMarie:
Build fixed and tests added.

marmarek:
the (new) test failed

marmarek (approved):

PR #159, opened 2024-05-04
Add support for exiting on stdin or stdout EOF     

DemiMarie:
This adds two new boolean service configuration options:
    
- exit-on-stdout-eof: exit when the socket service shuts down its output stream for writing.
- exit-on-stdin-eof: exit when the client sends EOF.

To avoid compatibility problems, global variables are used to pass this information from the configuration parser to the I/O code.  In the future, there should be a better way to pass this information, but this is not possible without more extensive changes.  Fortunately, the configuration parser only runs once in the life of any process right now, so this commit adds assertions to check this.  Qubes OS ships with assertions enabled, so any violation of this rule will be detected.
    
The main use of these features is to emulate the old qubes.ConnectTCP and qubes.UpdatesProxy services, which already had this behavior due to the use of socat.
    
These features are only supported for socket-based services, as executable services are more complicated and do not have a use case right now.
    
Currently, if a service exits due to exit-on-stdin-eof, the empty MSG_DATA_STDOUT that indicates EOF is not sent.  This is not a problem because qrexec-client-vm interprets MSG_DATA_EXIT_CODE as also indicating EOF on stdout and stderr.
    
Fixes: QubesOS/qubes-issues#9176

### DIFF
--- a/agent/qrexec-agent-data.c
+++ b/agent/qrexec-agent-data.c
@@ -283,8 +283,12 @@ static int handle_new_process_common(
 
     exit_code = process_io(&req);
 
-    if (type == MSG_EXEC_CMDLINE)
-        LOG(INFO, "pid %d exited with %d", pid, exit_code);
+    if (type == MSG_EXEC_CMDLINE) {
+        if (pid > 0)
+            LOG(INFO, "pid %d exited with %d", pid, exit_code);
+        else
+            LOG(INFO, "Socket service closed");
+    }
 
     libvchan_close(data_vchan);
     return exit_code;
--- a/agent/qrexec-agent-data.c
+++ b/agent/qrexec-agent-data.c
@@ -322,7 +322,7 @@ pid_t handle_new_process(int type, int connect_domain, int connect_port,
 /* Returns exit code of remote process */
 int handle_data_client(
     int type, int connect_domain, int connect_port,
-    int stdin_fd, int stdout_fd, int stderr_fd, int buffer_size, pid_t pid,
+    int stdin_fd, int stdout_fd, int buffer_size, pid_t pid,
     const char *extra_data)
 {
     int exit_code;
@@ -362,7 +362,7 @@ int handle_data_client(
 
     req.stdin_fd = stdin_fd;
     req.stdout_fd = stdout_fd;
-    req.stderr_fd = stderr_fd;
+    req.stderr_fd = -1;
     req.local_pid = pid;
 
     req.is_service = false;
--- a/agent/qrexec-agent-data.c
+++ b/agent/qrexec-agent-data.c
@@ -281,7 +281,7 @@ static int handle_new_process_common(
     req.prefix_data.data = NULL;
     req.prefix_data.len = 0;
 
-    exit_code = process_io(&req);
+    exit_code = qrexec_process_io(&req, cmd);
 
     if (type == MSG_EXEC_CMDLINE) {
         if (pid > 0)
@@ -382,7 +382,7 @@ int handle_data_client(
         req.prefix_data.len = 0;
     }
 
-    exit_code = process_io(&req);
+    exit_code = qrexec_process_io(&req, NULL);
     libvchan_close(data_vchan);
     return exit_code;
 }
--- a/libqrexec/exec.c
+++ b/libqrexec/exec.c
@@ -327,6 +327,8 @@ static int load_service_config_raw(struct qrexec_parsed_command *cmd,
                         config_full_path, sizeof(config_full_path), NULL);
     if (ret == -1)
         return 0;
+    if (ret != 0)
+        return ret;
     return qubes_toml_config_parse(config_full_path, &cmd->wait_for_session, user,
                                    &cmd->send_service_descriptor,
                                    &cmd->exit_on_stdout_eof,
--- a/libqrexec/exec.c
+++ b/libqrexec/exec.c
@@ -328,7 +328,16 @@ static int load_service_config_raw(struct qrexec_parsed_command *cmd,
     if (ret == -1)
         return 0;
     return qubes_toml_config_parse(config_full_path, &cmd->wait_for_session, user,
-                                   &cmd->send_service_descriptor);
+                                   &cmd->send_service_descriptor,
+                                   &cmd->exit_on_stdout_eof,
+                                   &cmd->exit_on_stdin_eof);
+}
+
+bool qrexec_cmd_use_fork_server(const struct qrexec_parsed_command *cmd) {
+    if (cmd == NULL)
+        return false;
+    return !cmd->nogui && cmd->send_service_descriptor && !cmd->exit_on_stdin_eof &&
+           !cmd->exit_on_stdout_eof;
 }
 
 int load_service_config_v2(struct qrexec_parsed_command *cmd) {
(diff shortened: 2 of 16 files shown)


DemiMarie:
Is there a way to mark some lines as “don’t expect to have coverage here”?  The lines that don’t have coverage are things like “junk data protocol version” (but that is caught long before) or “program started with one of its standard streams closed” (but that means a bug in the caller).

libqrexec/exec.c
```diff
@@ -311,9 +311,15 @@ static int find_file(
...
 }
 
+bool exit_on_stdout_eof = false;
+bool exit_on_stdin_eof = false;
+
 static int load_service_config_raw(struct qrexec_parsed_command *cmd,
                                    char **user)
 {
+    static bool called_yet = false;
+    assert(!called_yet);
```
marmarek, line 321: This takes down the whole qrexec-agent service. openqa says everything is broken...

```
[2024-05-04 02:28:41] [   35.640158] systemd[1]: Started systemd-coredump@0-800-0.service - Process Core Dump (PID 800/UID 0).
[2024-05-04 02:28:41] [   35.643650] audit: type=1130 audit(1714804121.207:110): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=systemd-coredump@0-800-0 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
[2024-05-04 02:28:41] [   35.996565] systemd-coredump[801]: [0;1;31mProcess 634 (qrexec-agent) of user 0 dumped core.
[2024-05-04 02:28:41] 
[2024-05-04 02:28:41] Module libeconf.so.0 from rpm libeconf-0.5.2-2.fc39.x86_64
[2024-05-04 02:28:41] Module libaudit.so.1 from rpm audit-3.1.2-8.fc39.x86_64
[2024-05-04 02:28:41] Module libpam.so.0 from rpm pam-1.5.3-3.fc39.x86_64
[2024-05-04 02:28:41] Module libvchan-xen.so.1 from rpm qubes-libvchan-xen-4.2.3-1.fc39.x86_64
[2024-05-04 02:28:41] Module libqrexec-utils.so.3 from rpm qubes-core-qrexec-4.2.18-1.2.fc39.x86_64
[2024-05-04 02:28:41] Module qrexec-agent from rpm qubes-core-qrexec-vm-4.2.18-1.2.fc39.x86_64
[2024-05-04 02:28:41] Stack trace of thread 634:
[2024-05-04 02:28:41] #0  0x00007402f0cbf834 __pthread_kill_implementation (libc.so.6 + 0x90834)
[2024-05-04 02:28:41] #1  0x00007402f0c6d8ee raise (libc.so.6 + 0x3e8ee)
[2024-05-04 02:28:41] #2  0x00007402f0c558ff abort (libc.so.6 + 0x268ff)
[2024-05-04 02:28:41] #3  0x00007402f0c5581b __assert_fail_base.cold (libc.so.6 + 0x2681b)
[2024-05-04 02:28:41] #4  0x00007402f0c65c57 __assert_fail (libc.so.6 + 0x36c57)
[2024-05-04 02:28:41] #5  0x00007402f0e2f664 load_service_config_raw (libqrexec-utils.so.3 + 0x6664)
[2024-05-04 02:28:41] #6  0x00007402f0e2f7bf load_service_config_v2 (libqrexec-utils.so.3 + 0x67bf)
[2024-05-04 02:28:41] #7  0x00005eadb2cf3293 main (qrexec-agent + 0x4293)
[2024-05-04 02:28:41] #8  0x00007402f0c5714a __libc_start_call_main (libc.so.6 + 0x2814a)
[2024-05-04 02:28:41] #9  0x00007402f0c5720b __libc_start_main@@GLIBC_2.34 (libc.so.6 + 0x2820b)
[2024-05-04 02:28:41] #10 0x00005eadb2cf3d25 _start (qrexec-agent + 0x4d25)
[2024-05-04 02:28:41] ELF object binary architecture: AMD x86-64
[2024-05-04 02:28:41] [0m
[2024-05-04 02:28:41] [   36.000765] audit: type=1131 audit(1714804121.565:111): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=systemd-coredump@0-800-0 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
[2024-05-04 02:28:41] [   36.002341] systemd[1]: systemd-coredump@0-800-0.service: Deactivated successfully.
[2024-05-04 02:28:41] [   36.024073] systemd[1]: [0;1;38;5;185mqubes-qrexec-agent.service: Main process exited, code=dumped, status=6/ABRT[0m
[2024-05-04 02:28:41] [   36.024239] systemd[1]: [0;1;38;5;185mqubes-qrexec-agent.service: Failed with result 'core-dump'.[0m
```

maybe set initial value to false here explicitly, instead of assuming the config is loaded only once per process?

marmarek:
> Is there a way to mark some lines as “don’t expect to have coverage here”?

I don't know, but sounds useful to have. Maybe gcov documentation says something about this? Or codecov one?

agent/qrexec-agent-data.c
```diff
@@ -281,9 +281,9 @@ static int handle_new_process_common(
     req.prefix_data.data = NULL;
     req.prefix_data.len = 0;
 
-    exit_code = process_io(&req);
+    exit_code = qrexec_process_io(&req, cmd);
 
-    if (type == MSG_EXEC_CMDLINE)
+    if (type == MSG_EXEC_CMDLINE && pid > 0)
         LOG(INFO, "pid %d exited with %d", pid, exit_code);
```
marmarek, line 287: Can I get some log message on service exit?

libqrexec/exec.c
```diff
@@ -731,6 +742,18 @@ int find_qrexec_service(
                 path_buffer.data);
             return -2;
         }
+        if (cmd->exit_on_stdout_eof) {
+            LOG(ERROR, "Refusing to execute executable service %s with "
+                       "exit-on-service-eof=true",
```
marmarek, line 747: This is not the config name actually used.

libqrexec/exec.c
```diff
@@ -731,6 +742,18 @@ int find_qrexec_service(
...
+                path_buffer.data);
+            return -2;
+        }
+        if (cmd->exit_on_stdin_eof) {
+            LOG(ERROR, "Refusing to execute executable service %s with "
+                       "exit-on-client-eof=true",
```
marmarek, line 753: And here

marmarek (changes requested):
"Implement new version of process_io() that takes extensible struct" are partially included in "Add support for exiting on stdin or stdout EOF" - specifically the second iteration of process_io refactor. And also the commit message of the latter is outdated now - the global variables thing is not true.

agent/qrexec-agent-data.c
DemiMarie: I think that should be a separate commit, since the current log message is completely unhelpful.

DemiMarie (commented):
Which name should I use in the config?  `exit-on-stdin-eof`/`exit-on-stdout-eof` or `exit-on-client-eof`/`exit-on-server-eof`?

agent/qrexec-agent-data.c
marmarek: It is helpful in that journald adds the logging process PID to the message, which can be easily correlated with the service start log.

marmarek:
> Which name should I use in the config? `exit-on-stdin-eof`/`exit-on-stdout-eof` or `exit-on-client-eof`/`exit-on-server-eof`?

That's a good question. Both require some documentation to clarify the behavior, but maybe the latter is a bit clearer? idk...

DemiMarie:
> > Which name should I use in the config? `exit-on-stdin-eof`/`exit-on-stdout-eof` or `exit-on-client-eof`/`exit-on-server-eof`?
> 
> That's a good question. Both require some documentation to clarify the behavior, but maybe the latter is a bit clearer? idk...

I think so too.  I kept using stdin/stdout internally, though, because that is what the file descriptors are named and so it makes the code easier to read.

marmarek (approved):

PR #162, opened 2024-05-16
qrexec-daemon: Use close_range where possible

DemiMarie:
This makes things faster and trivially more correct.


marmarek:
PipelineRetry

marmarek (approved):

PR #167, opened 2024-05-28
Minor improvements

marmarek:


marmarek:
@DemiMarie can you review?

libqrexec/log.c
```diff
@@ -88,5 +88,8 @@ void setup_logging(const char *program_name) {
             errx(125, "File descriptor %d is closed, cannot continue", i);
     }
 
+    /* qrexec_logv() uses several separate fprintf calls, print them at once */
+    setlinebuf(stderr);
```
DemiMarie, line 92:
```suggestion
    if (setvbuf(stderr, NULL, _IOLBF, 0))
        errx(1, "failed to enable line buffering on stderr");
```

`setvbuf()` is the C11 and POSIX standard interface, `setlinebuf()` is presumably a legacy alias.

daemon/qrexec-daemon-common.c
```diff
@@ -135,7 +135,8 @@ bool qrexec_execute_vm(const char *target, bool autostart, int remote_domain_id,
             return false;
         }
     }
-    if (autostart) {
+    int s = connect_unix_socket(target);
+    if (s == -2 && autostart) {
```
DemiMarie, line 139: This should only be done if the target VM is not a disposable VM.  If the target VM is a fresly created disposable VM, it will definitely not be running here, so the check is pointless.
```suggestion
    int s = disposable ? -2 : connect_unix_socket(target);
    if (s == -2 && autostart) {
```

DemiMarie (changes requested):

libqrexec/log.c
marmarek: Do you think failure here really should be fatal?

DemiMarie: It doesn’t strictly need to be, and if there was any reasonable reason for this function to fail I would agree that it should not be fatal.  However, the only reasons I can think of are either:

1. libc is buggy.
2. libc cannot allocate an 8KiB buffer.

In both cases, something is already horribly wrong.

That said, I would not oppose changing this to a warning.

marmarek: > libc cannot allocate an 8KiB buffer.

according to documentation, the buffer is allocated only on first print, not here

DemiMarie (commented):
Nit

libqrexec/log.c
```diff
@@ -88,5 +88,9 @@ void setup_logging(const char *program_name) {
             errx(125, "File descriptor %d is closed, cannot continue", i);
     }
 
+    /* qrexec_logv() uses several separate fprintf calls, print them at once */
+    if (setvbuf(stderr, NULL, _IOLBF, 0))
+        warn("Failed to enable line buffering on stderr");
```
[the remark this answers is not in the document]
marmarek: It says "It may set errno on failure", so better log it in case it has something useful.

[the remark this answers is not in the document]
DemiMarie: Maybe set `errno` to 0 first then?  That way, if errno is _not_ set, it is not some stale value that could be actively misleading.

DemiMarie (approved):
Fine to merge as-is.  I have a suggestion (`errno = 0` before `setvbuf`) but it is up to you whether to take it or leave it.

DemiMarie (approved):

PR #169, opened 2024-06-19
Extra internal validation

DemiMarie:
This adds some extra internal validation.


marmarek:
CI says you broke legitimate case here...

marmarek:
PipelineRetry

marmarek:
Tests fail. And formatting need an update too

PR #170, opened 2024-06-27
policy-daemon: invalidate cache on file move too

marmarek:
Invalidate cache not only when file is modified inside policy dir, but
also when it's moved in or out.

### DIFF
--- a/qrexec/tests/policy_cache.py
+++ b/qrexec/tests/policy_cache.py
@@ -98,6 +98,47 @@ class TestPolicyCache:
         assert not cache.outdated
 
     @pytest.mark.asyncio
+    async def test_14_policy_move(self, tmp_path, mock_parser):
+        policy_path = tmp_path / "policy"
+        policy_path.mkdir()
+        cache = utils.PolicyCache(policy_path)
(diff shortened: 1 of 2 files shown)


DemiMarie:
Looks geat, except there are no tests :slightly_smiling_face:.

qrexec/tests/policy_cache.py
```diff
@@ -97,6 +97,47 @@ async def test_13_no_change(self, tmp_path, mock_parser):
...
+        file_moved.rename(file)
+
+        await asyncio.sleep(1)
+
+        assert cache.outdated
+
+        cache.get_policy()
+
+        call = unittest.mock.call(policy_path=policy_path)
+        assert mock_parser.mock_calls == [call, call, call]
```
DemiMarie, line 139: What about moving within the directory?

DemiMarie (commented):
One more test that could be added, but IIUC it won’t add new coverage.

qrexec/tests/policy_cache.py
marmarek: It's then both move from and move to, so this case is covered too.

PR #171, opened 2024-07-01
Support substring matching in qube name text fields

alimirjamali:
fixes: https://github.com/QubesOS/qubes-issues/issues/8583

### DIFF
--- a/qrexec/tools/qrexec_policy_agent.py
+++ b/qrexec/tools/qrexec_policy_agent.py
@@ -184,11 +184,22 @@ class VMListModeler:
                 completion.set_inline_selection(True)
                 completion.set_inline_completion(True)
                 completion.set_popup_completion(True)
-                completion.set_popup_single_match(False)
+                completion.set_popup_single_match(True)
                 completion.set_model(list_store)
                 completion.set_text_column(1)
 
                 entry_box.set_completion(completion)
+
+                def qube_matching_function(completion: Gtk.EntryCompletion,
+                                           key: str,
+                                           iterator: Gtk.TreeIter,
+                                           user_data: object) -> bool:
+                    # pylint: disable=unused-argument
+                    modelstr = completion.get_model()[iterator][1]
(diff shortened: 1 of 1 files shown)


alimirjamali:
Making Pylint happy(er)

marmarta (approved):

marmarta:
PipelineRetryFailed

alimirjamali:
Is this related to this commit? Or a policy error?

image

marmarek:
The other qrexec PR in the batch is broken. Now rerunning tests without it.

qrexec/tools/qrexec_policy_agent.py
```diff
@@ -184,11 +184,22 @@ def apply_model(
                 completion.set_inline_selection(True)
                 completion.set_inline_completion(True)
                 completion.set_popup_completion(True)
-                completion.set_popup_single_match(False)
+                completion.set_popup_single_match(True)
```
marmarek, line 187: Can you explain this change? It means you can't accept a single match (after typing it in) with a single click anymore, you need two clicks. OpenQA found this out, but TBH I'm not sure if that's a common case.

alimirjamali: > Can you explain this change?

That is something I found only after using the patch for a while. The default logic (matching from the start) does not require to show the popup for a single match if user has entered the entire string.

In our case, if user wants to search for `disp1234` and user has currenty `disp1234`, `dips5678` and `disp90ab` running at the moment and user types `1`, there will not popup result to select `disp1234` from it if `popup_single_match` is et to False.

alimirjamali: p.s. Actually the GTK default matching algorithm is also insufficient if a popup is not shown. As it is case in-senstivie and Qube names are case-sensitive.

marmarek: I see, ok.

PR #174, opened 2024-09-13
tests: fix POLICY_PROGRAM and update documentation

fepitre:


qrexec/tests/socket/daemon.py
```diff
@@ -49,9 +50,27 @@ class TestDaemon(unittest.TestCase):
...
-exit $(cat {tempdir}/qrexec-policy-exitcode || echo 1)
+exit_code=$(cat {tempdir}/qrexec-policy-exitcode || echo 1)
+
+# Prepare the response based on the exit code
+if [ "$exit_code" -eq 0 ]; then
+    # Allow response
+    printf 'result=allow\n'
+    printf 'autostart=True\n'
+    printf 'user=toto\n'
+    printf 'target=%s\n' "$2"
```
marmarek, line 65: $2 is the source, not target in the 5-arg version (that qrexec-daemon will call)

qrexec/tests/socket/daemon.py
```diff
@@ -268,6 +287,58 @@ def recv_refused(agent):
...
+requested_target=anotherdomain
+"""
+
+        # check allowed request
+        agent.send_message(
+            qrexec.MSG_TRIGGER_SERVICE3,
+            struct.pack("<64s32s", target_domain_name.encode(), ident.encode())
+            + b"a\0",
+        )
+        message_type, data = agent.recv_message()
```
marmarek, line 334: I guess this only works because your policy redirected the request back to the source domain (see earlier comment). For this to really work, you should have `target_domain_name = self.domain_name` (real policy will refuse loopback connections, but fortunately you don't have real policy here :) ).

marmarek (changes requested):

qrexec/tests/socket/daemon.py
fepitre: Yes I forgot to ask you, and you answered in the comment below about using self.domain_name instead.

marmarek (approved):

PR #176, opened 2024-09-23
Provide clear error on invalid policy file name

alimirjamali:
Improves: https://github.com/QubesOS/qubes-issues/issues/8795

### DIFF
--- a/qrexec/policy/admin.py
+++ b/qrexec/policy/admin.py
@@ -24,6 +24,7 @@ from grp import getgrnam
 import contextlib
 import fcntl
 import os
+import re
 import hashlib
 
 from .parser import ValidateParser, FilePolicy, get_invalid_characters
@@ -87,7 +88,11 @@ class PolicyAdmin:
         Throws PolicyAdminException in case of user error.
         """
 
-        assert all(char in RPCNAME_ALLOWED_CHARSET for char in arg)
+        if not all(char in RPCNAME_ALLOWED_CHARSET for char in arg):
+            raise PolicyAdminException(
+                "Invalid argument: \"{}\"\n"
+                "Valid characters are letters, numbers, dot, plus, hyphen and "
+                "underline".format(arg))
 
         func = self._find_method(service_name)
         if not func:
@@ -266,7 +271,12 @@ class PolicyAdmin:
 
     # helpers
 
-    def _get_path(self, arg: str, dir_path, suffix: str) -> Path:
+    def _get_path(self, arg: str, dir_path: str , suffix: str) -> Path:
+        if not re.compile(r'^[\w-]+$').match(arg):
+            raise PolicyAdminException(
+                f"Invalid policy file name: {arg}\n"
+                "Names must contain only alphanumeric characters, "
+                "underscore and hyphen.")
         path = dir_path / (arg + suffix)
         path = path.resolve()
         if path.parent != dir_path:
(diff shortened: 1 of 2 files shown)


ben-grande:
The RPC name and the filename have slightly different characters allowed. It doesn't seem to be allowed to use uppercase letters in the filename, this could impact sorting on different locales.

https://github.com/QubesOS/qubes-core-qrexec/blob/7d66382ec544b418badc9b47362d2fcb251567c9/qrexec/policy/parser.py#L58

https://github.com/QubesOS/qubes-core-qrexec/blob/7d66382ec544b418badc9b47362d2fcb251567c9/qrexec/policy/parser.py#L85-L88

---

Thanks for the contributions.

alimirjamali:
Thanks for the review

> The RPC name and the filename have slightly different characters allowed.

So it might be appropriate to ignore deviation here. Fixing the GUI front-end Regex is definitely needed. And small improvement for the error message of RPC won't be bad.

qrexec/policy/admin.py
```diff
@@ -87,7 +87,13 @@ def handle_request(
         Throws PolicyAdminException in case of user error.
         """
 
-        assert all(char in RPCNAME_ALLOWED_CHARSET for char in arg)
+        try:
+            assert all(char in RPCNAME_ALLOWED_CHARSET for char in arg)
+        except AssertionError as ex:
```
marmarek, line 92: If changing to a different exception, I'd prefer simply `if not all(....): raise` instead of converting one exception to another.

Anyway, as @ben-grande noticed, this isn't exactly correct. This verifies call argument, but depending on specific call, it may be a file name (most of the methods here), or a service name (at least `policy.GetFiles`). The latter one already verifies its argument separately, maybe better to include also file name verification (with appropriate error message) in the `_get_path` function with rules guarding file names?

marmarek (changes requested):

qrexec/policy/admin.py
alimirjamali: > maybe better to include also file name verification (with appropriate error message) in the `_get_path` function with rules guarding file names?

There is a slight problem with this approach. `handle_request` and its validation happens before `_get_path` function call. It may be better to implement both type of validation within `handle_request` itself. I will submit the amended PR shortly.

alimirjamali: @marmarek 
I believe this is ready (including updated unittests). The failed test appears to be due to a temporal network connection. I do not have access to RetryPipelineFailed command.

marmarek: > `handle_request` and its validation happens before `_get_path` function call.

Technically, it's only partially true. The final part of `handle_request` calls the actual handler that internally calls `_get_path`. So, as long as validation in `handle_request` itself is not stricter than one needed in `_get_path`, it should be okay.

qrexec/policy/admin.py
```diff
@@ -87,7 +88,17 @@ def handle_request(
         Throws PolicyAdminException in case of user error.
         """
 
-        assert all(char in RPCNAME_ALLOWED_CHARSET for char in arg)
+        if service_name in ["policy.Get", "policy.Replace", "policy.Remove"] \
```
marmarek, line 91: This missed a few methods that also take file name (all the "include" ones). Anyway, as said in the other comment, it doesn't necessarily needs to be in this place.

alimirjamali: OK. I will move it to `_get_path`. But the validation within `handle_request` will definitely precede that. Overall this should not be a serious issue since (the fixed) validation within `qubes-policy-editor-gui` will precede all of them.

marmarek (approved):

PR #179, opened 2024-09-30
Pipe `qrexec-client(-vm) --help` output to stdout

alimirjamali:
Help is output to stderr at the moment

### DIFF
--- a/agent/qrexec-client-vm.c
+++ b/agent/qrexec-client-vm.c
@@ -112,19 +112,21 @@ static struct option longopts[] = {
 };
 
 _Noreturn static void usage(const char *argv0, int status) {
-    fprintf(stderr,
+    FILE *stream = status ? stderr : stdout;
+    fprintf(stream,
             "usage: %s [options] target_vmname program_ident [local_program [local program arguments]]\n",
             argv0);
-    fprintf(stderr, "Options:\n");
-    fprintf(stderr, "  -h, --help - print this message\n");
-    fprintf(stderr, "  --buffer-size=BUFFER_SIZE - minimum vchan buffer size (default: 64k)\n");
-    fprintf(stderr, "  -t, --filter-escape-chars-stdout - filter non-ASCII and control characters on stdout (default if stdout is a terminal)\n");
-    fprintf(stderr, "  -T, --filter-escape-chars-stderr - filter non-ASCII and control characters on stderr (default if stderr is a terminal)\n");
-    fprintf(stderr, "  --no-filter-escape-chars-stdout - opposite to --filter-escape-chars-stdout\n");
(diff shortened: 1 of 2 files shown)


alimirjamali:
Notes: I can not locate the proper unittest to update. Guidance is appreciated.

daemon/qrexec-client.c
```diff
@@ -174,8 +175,11 @@ int main(int argc, char **argv)
     int rc = QREXEC_EXIT_PROBLEM;
 
     if (argc < 3) {
+        if (argc == 2 && (strcmp(argv[1], "-h") == 0 || strcmp(argv[1], "--help") == 0)) {
```
marmarek, line 178: Better move the check after the getopt_long loop (or maybe even drop it completely, since there is already `optind >= argc` check below), instead of duplicating --help handling.

agent/qrexec-client-vm.c
```diff
@@ -112,19 +112,20 @@ static struct option longopts[] = {
...
-    fprintf(stderr, "Options:\n");
-    fprintf(stderr, "  -h, --help - print this message\n");
-    fprintf(stderr, "  --buffer-size=BUFFER_SIZE - minimum vchan buffer size (default: 64k)\n");
-    fprintf(stderr, "  -t, --filter-escape-chars-stdout - filter non-ASCII and control characters on stdout (default if stdout is a terminal)\n");
-    fprintf(stderr, "  -T, --filter-escape-chars-stderr - filter non-ASCII and control characters on stderr (default if stderr is a terminal)\n");
-    fprintf(stderr, "  --no-filter-escape-chars-stdout - opposite to --filter-escape-chars-stdout\n");
-    fprintf(stderr, "  --no-filter-escape-chars-stderr - opposite to --filter-escape-chars-stderr\n");
-    fprintf(stderr, "  --agent-socket=PATH - path to connect to, default: %s\n", QREXEC_AGENT_TRIGGER_PATH);
-    fprintf(stderr, "  -p PREFIX-DATA, --prefix-data=PREFIX-DATA - send the given data before the provided stdin (can only be used once)\n");
-    fprintf(stderr, "  --use-stdin-socket - use fd 0 (which must be socket) for both stdin and stdout\n");
```
marmarek, line 127: What happened to the `--use-stdin-socket` option?

marmarek (changes requested):

agent/qrexec-client-vm.c
alimirjamali: Ops. Will fix it

daemon/qrexec-client.c
alimirjamali: OK

marmarek (approved):

PR #182, opened 2024-10-15
Support for source domain in RPC service

fepitre:
QubesOS/qubes-issues/issues/9475
QubesOS/qubes-issues/issues/9015


run-tests
```diff
@@ -1,8 +1,8 @@
-#!/bin/bash -e
+#!/bin/bash -ex
 
 case $0 in (/*) cd "${0%/*}/";; (*/*) cd "./${0%/*}";; esac
 if command -v dnf >/dev/null; then
-    sudo dnf install python3dist\({coverage,pytest,gbulb,pyinotify,pytest-asyncio}\) || :
+    sudo dnf install python3dist\({coverage,pytest,gbulb,pyinotify,pytest-asyncio}\) qubes-libvchan-xen-devel pam-devel || :
```
fepitre, line 5: Keeping it here for now to not forget packages to install ..

daemon/qrexec-daemon-common.c
```diff
@@ -84,7 +84,7 @@ int handle_daemon_handshake(int fd)
         return -1;
     }
 
-    if (info.version != QREXEC_PROTOCOL_VERSION) {
+    if (info.version < QREXEC_PROTOCOL_V3) {
```
marmarek, line 87: Generally, it's not enough to just relax this check. You need to also save which version the daemon is using (V4 is just adding a message that isn't relevant for the qrexec-client, but still). And similarly change in `handle_client_hello` and also save which client is using what version.

Alternatively, you can keep using V3 for daemon<->client communication, since new message is not needed at this place (right?). This will be also easier to test. But relaxing the check is a good idea for the future anyway (but can be done in a separate PR if you prefer).

marmarek (changes requested):

marmarek:
The update fails due to mismatch between qrexec-client and qrexec-daemon versions for already running VMs:
```
2025-03-18 14:39:37.659 qrexec-daemon[2760]: qrexec-daemon.c:732:handle_client_hello: Incompatible client protocol version (remote 4, local 3)
2025-03-18 14:39:38.162 qrexec-daemon[2760]: qrexec-daemon.c:732:handle_client_hello: Incompatible client protocol version (remote 4, local 3)
2025-03-18 14:39:38.866 qrexec-daemon[2760]: qrexec-daemon.c:732:handle_client_hello: Incompatible client protocol version (remote 4, local 3)
```
See my earlier comment here about possible solution.

qrexec/tests/socket/qrexec.py
```diff
@@ -93,12 +94,14 @@ def recv_all_messages(self):
             messages.append((message_type, data))
         return messages
 
-    def handshake(self):
-        self.send_message(MSG_HELLO, struct.pack("<L", QREXEC_PROTOCOL_VERSION))
+    def handshake(self, protocol_version=None):
+        if not protocol_version:
+            protocol_version = QREXEC_CLIENT_PROTOCOL_VERSION
```
marmarek, line 99: Wouldn't it make more sense to do the other way around? Default to version QREXEC_PROTOCOL_VERSION (set as 4?), only in cases where older version is required provide the parameter. I think you'll need to use the parameter in less places then.

And also, there is no need for `=None` and then the if - simply put default value in the function header.

fepitre: Yes multiple iterations and removing of code. I did it the way around.

marmarek:
Something went wrong with version negotiation between qrexec-daemon and qrexec-agent. With new daemon in dom0 and old agent in VM all is fine. But after updating agent in VM it fails:
```
2025-03-27 10:41:39.121 qrexec-daemon[12796]: qrexec-daemon.c:1773:main: qrexec-agent has disconnected
2025-03-27 10:41:39.178 qrexec-daemon[12796]: qrexec-daemon.c:1638:handle_agent_restart: qrexec-agent has reconnected
2025-03-27 10:41:42.871 qrexec-daemon[12796]: qrexec-daemon.c:1354:sanitize_message_from_agent: agent sent (new) MSG_TRIGGER_SERVICE4 although it uses protocol 3
2025-03-27 10:42:18.018 qrexec-daemon[12984]: process_io.c:250:qrexec_process_io: vchan connection closed early (fds: 4 4 -1, status: -1 -1)
```

Hopefully https://github.com/QubesOS/qubes-core-qrexec/pull/182 will fix it (and hopefully it won't result in merge conflict...

qrexec/tests/socket/agent.py
```diff
@@ -181,11 +180,10 @@ def connect_client(self):
 
 @unittest.skipIf(os.environ.get("SKIP_SOCKET_TESTS"), "socket tests not set up")
 class TestAgent(TestAgentBase):
+
     def test_handshake(self):
         self.start_agent()
 
-        dom0 = self.connect_dom0()
```
marmarek, line 187: This looks like unintentional removal

daemon/qrexec-daemon.c
```diff
@@ -980,16 +995,19 @@ struct QrexecPolicyRequest {
...
     ssize_t bytes_sent = 0;
     int command_size = asprintf(&command,
             "source=%s\n"
+            "relayed_source=%s\n"
             "intended_target=%s\n"
             "service_and_arg=%s\n\n",
             remote_domain_name,
+            source_domain,
             target_domain,
             service_name);
```
marmarek, line 1012: String manipulation in C is horrible, but you can do like this:
```suggestion
    int command_size = asprintf(&command,
            "source=%s"
            "%s%s"
            "\nintended_target=%s"
            "\nservice_and_arg=%s"
            "\n\n",
            remote_domain_name,
            source_domain ? "\nrelayed_source=" : "",
            source_domain ? : "",
            target_domain,
            service_name);
```

agent/qrexec-client-vm.c
```diff
@@ -250,14 +258,16 @@ int main(int argc, char **argv)
...
     trigger_fd = connect_unix_socket(agent_trigger_path);
 
-    hdr.type = MSG_TRIGGER_SERVICE3;
+    hdr.type = MSG_TRIGGER_SERVICE4;
     hdr.len = sizeof(params) + service_name_len;
 
     memset(&params, 0, sizeof(params));
 
+    if (source_qube!=NULL)
+        strncpy(params.source_domain, source_qube, sizeof(source_qube) - 1);
```
marmarek, line 267:
```suggestion
        strncpy(params.source_domain, source_qube, sizeof(params.source_domain) - 1);
```
First of all, when copying string, limit copy size to the _target_ space, otherwise any limitation wouldn't prevent overflow. But then, `sizeof(char*)` is 8, so also not what you want.

marmarek:
Remapping didn't work:
```
Apr 04 17:26:02.484189 dom0 qrexec-policy-daemon[1718]: qrexec: qubes.Filecopy+: test-inst-vm1 -> test-inst-vm2-remote: allowed to test-inst-vm2-remote
Apr 04 17:28:38.216407 dom0 qrexec-policy-daemon[1718]: qrexec: qubes.Filecopy+: test-inst-relay -> test-inst-vm2: denied: denied by the user /etc/qubes/policy.d/90-default.policy:31
```
The second line should be test-inst-vm1-remote -> test-inst-vm2.

agent/qrexec-client-vm.c
```diff
@@ -127,14 +128,15 @@ _Noreturn static void usage(const char *argv0, int status) {
             QREXEC_AGENT_TRIGGER_PATH);
     fprintf(stream, "  -p PREFIX-DATA, --prefix-data=PREFIX-DATA - send the given data before the provided stdin (can only be used once)\n");
     fprintf(stream, "  --use-stdin-socket - use fd 0 (which must be socket) for both stdin and stdout\n");
+    fprintf(stderr, "  -s SOURCE-QUBE, --source-qubes=SOURCE-QUBE - Specify the source qube. This option provides extra information for relayed connection that may be used during policy evaluation.\n");
```
marmarek, line 131:
```suggestion
    fprintf(stderr, "  -s SOURCE-QUBE, --source-qube=SOURCE-QUBE - Specify the source qube. This option provides extra information for relayed connection that may be used during policy evaluation.\n");
```

marmarek (changes requested):

marmarek:
This is a good description of the whole process, but there should be also a section with just the information how transport rpc needs to be written. This needs to include specific information what such service gets and how (call arguments, env variables, qubesdb entries etc). And maybe a simple example with explanation of each step?

fepitre:
> This is a good description of the whole process, but there should be also a section with just the information how transport rpc needs to be written. This needs to include specific information what such service gets and how (call arguments, env variables, qubesdb entries etc). And maybe a simple example with explanation of each step?

Yes this is to be added. I can add either SSH example or a dummy version (understanding here on the same host) based on the core admin test?

marmarek:
On Fri, Apr 18, 2025 at 12:46:18AM -0700, Frédéric Pierret wrote:
> Yes this is to be added. I can add either SSH example or a dummy version (understanding here on the same host) based on the core admin test? 

The SSH one is probably better, the dummy is confusing due to having the
same host playing two roles.

-- 
Best Regards,
Marek Marczykowski-Górecki
Invisible Things Lab

doc/qrexec-remotevm.rst
```diff
@@ -0,0 +1,173 @@
...
+
+   if [[ -z "$remote_qube" ]]; then
+       echo "Error: Could not read remote qube for target '$target'."
+       exit 1
+   fi
+
+   # Forward the qrexec call via SSH
+   ssh "$remote_qube" qrexec-client-vm --source-qube="$QREXEC_REMOTE_DOMAIN" "$remote_qube" "$service"
+
+Workflow Overview
```
marmarek, line 146: IMO better to put description before the example. The description is (should be) what the service should do, and then example based on SSH. It should be possible to write also other services based on this description.

doc/qrexec-remotevm.rst
```diff
@@ -0,0 +1,173 @@
...
+"""""""""""""""""
+
+1. **Local-Qube to Local-Relay**: A Qubes RPC call is made from Local-Qube to Local-Relay with arguments of the form:
```
marmarek, line 149: Add info that this is what the transport_rpc property is about.

doc/qrexec-remotevm.rst
```diff
@@ -0,0 +1,173 @@
...
+
+   ``Remote-Qube+<service>``
+
+    where `<service>` contains the original service with argument that has been processed by policies.
+
+2. **Destination lookup**: The RPC script extracts `Remote-Qube` and queries QubesDB on Local‑Relay to verify and translate it into the actual qube name in `Remote-QubesOS` (in this example it is assumed to be the same).
```
marmarek, line 155: Include exact qubesdb path in this description.

doc/qrexec-remotevm.rst
```diff
@@ -0,0 +1,173 @@
...
+
+3. **SSH**: The script uses SSH to connect to the identified host (configured to forward to the Remote‑Relay). It then runs ``qrexec-client-vm`` on that host, passing along the original source domain via the environment variable ``$QREXEC_REMOTE_DOMAIN``.
+
+4. **Remote-QubesOS policy evaluation**: `Remote‑QubesOS` applies its policy rules for a request from `Local‑Relay` to `Remote‑Qube` with information that source qube is `Local-Qube`.
+
+SSH Client Configuration Hint
++++++++++++++++++++++++++++++
+
+To ensure that any SSH connection to a given host actually lands on the corresponding `Remote‑Relay`, add an entry like the following to your ``/etc/ssh/ssh_config``:
```
marmarek, line 164: Maybe better `~/.ssh/config`? It will persist without extra bind-dirs then.

As for the hints - on the remote site /etc/ssh better be made persistent with bind-dirs (and the host key re-generated there), otherwise host key will be shared across all VMs based on the same template, so not a secret anymore...

doc/qrexec-remotevm.rst
```diff
@@ -0,0 +1,173 @@
...
+   # Retrieve the remote Qube identifier
+   remote_qube="$(qubesdb-read "/remote/$target" 2>/dev/null || true)"
+
+   if [[ -z "$remote_qube" ]]; then
+       echo "Error: Could not read remote qube for target '$target'."
+       exit 1
+   fi
+
+   # Forward the qrexec call via SSH
+   ssh "$remote_qube" qrexec-client-vm --source-qube="$QREXEC_REMOTE_DOMAIN" "$remote_qube" "$service"
```
marmarek, line 144: Theoretically, `--source-qube="$QREXEC_REMOTE_DOMAIN"` on remote side may need a similar lookup via qubesdb (on remote side) too. But IMO it's ok to simplify it here, just leave a comment that using `QREXEC_REMOTE_DOMAIN` directly here assumes the remote name matches the local one.

doc/qrexec-remotevm.rst
```diff
@@ -0,0 +1,173 @@
...
+
+To ensure that any SSH connection to a given host actually lands on the corresponding `Remote‑Relay`, add an entry like the following to your ``/etc/ssh/ssh_config``:
+
+.. code-block:: ini
+
+   Host *
+       HostName <Remote-Relay_ip>
+
+Adjust the pattern, hostname, and jump host settings to match your environment.
+This guarantees that SSH connections intended for a RemoteVM are transparently proxied through Remote‑Relay./
```
marmarek, line 172: Stray `/` at the end.

marmarek (changes requested):

doc/qrexec-remotevm.rst
fepitre: I know but for me it's up to the user to know that's configuring host key at system level (ofc we would document it in qubes-doc at the end). Also for home path, is RPC using user home?

marmarek: Services are called as the default user normally, so yes, it will be `user`.

doc/qrexec-remotevm.rst
```diff
@@ -0,0 +1,175 @@
...
+Prerequisites
+^^^^^^^^^^^^^
+
+- SSH key-based authentication is configured between all involved hosts (Local‑Relay, Remote‑Relay, and any intermediate jump hosts).
+- (Optional) Configure any necessary proxies or jump hosts by following the OpenSSH Cookbook guidelines: Proxies and Jump Hosts <https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Proxies_and_Jump_Hosts>.
+
+Workflow Overview
+^^^^^^^^^^^^^^^^^
+
+1. **Local-Qube to Local-Relay**: The `transport_rpc` property of `Remote‑Relay` is used to refer to RPC call to be made from Local-Qube to Local-Relay with arguments of the form:
```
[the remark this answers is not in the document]
fepitre: none of them, `Local-Qube`

doc/qrexec-remotevm.rst
fepitre: Yes let's ignore with a warning.

marmarek (approved):

PR #185, opened 2024-12-30
docs: qrexec-client accepts user:command not command

zaz:
qrexec-client -d untrusted command

fails, but

    qrexec-client -d untrusted user:command

works.

We could also mention that `DEFAULT` is accepted for user.


marmarek:
> We could also mention that `DEFAULT` is accepted for user.

Yes, that makes sense

marmarek (approved):

PR #186, opened 2025-01-09
Do not close FDs 0, 1, or 2

DemiMarie:
If they are closed, another file descriptor could be created with these numbers, and so standard library functions that use them might write to an unwanted place.  dup2() a file descriptor to /dev/null over them instead.


DemiMarie:
Codecov appears to not be testing what happens in the child process after fork() and the error path of “cannot open `/dev/null`”.

marmarek:
AFAIR unit tests do not cover the PAM handling part, as they are not running as a system service, test runners don't have necessary PAM configuration etc.

marmarek:
> system_tests_qrexec
> 
>     * TC_00_Qrexec_debian-12-xfce: [test_053_qrexec_vm_service_eof_reverse](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_debian-12-xfce/4) (failure)
>       `AssertionError: Timeout, probably EOF wasn't transferred`
> 
>     * TC_00_Qrexec_debian-12-xfce: [test_092_qrexec_service_socket_dom0_eof_reverse](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_debian-12-xfce/18) (failure)
>       `AssertionError: service timeout, probably EOF wasn't transferred fr...`
> 
>     * TC_00_Qrexec_debian-12-xfce: [test_098_qrexec_service_socket_vm_eof](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_debian-12-xfce/23) (failure)
>       `AssertionError: service timeout, probably EOF wasn't transferred to...`
> 
>     * TC_00_Qrexec_fedora-41-xfce: [test_053_qrexec_vm_service_eof_reverse](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_fedora-41-xfce/4) (failure)
>       `AssertionError: Timeout, probably EOF wasn't transferred`
> 
>     * TC_00_Qrexec_fedora-41-xfce: [test_092_qrexec_service_socket_dom0_eof_reverse](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_fedora-41-xfce/18) (failure)
>       `AssertionError: service timeout, probably EOF wasn't transferred fr...`
> 
>     * TC_00_Qrexec_fedora-41-xfce: [test_098_qrexec_service_socket_vm_eof](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_fedora-41-xfce/23) (failure)
>       `AssertionError: service timeout, probably EOF wasn't transferred to...`
> 
>     * TC_00_Qrexec_whonix-gateway-17: [test_053_qrexec_vm_service_eof_reverse](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_whonix-gateway-17/4) (failure)
>       `AssertionError: Timeout, probably EOF wasn't transferred`
> 
>     * TC_00_Qrexec_whonix-gateway-17: [test_083_qrexec_service_argument_specific_implementation](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_whonix-gateway-17/14) (error)
>       `subprocess.CalledProcessError: Command '/usr/lib/qubes/qrexec-clien...`
> 
>     * TC_00_Qrexec_whonix-gateway-17: [test_092_qrexec_service_socket_dom0_eof_reverse](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_whonix-gateway-17/18) (failure)
>       `AssertionError: service timeout, probably EOF wasn't transferred fr...`
> 
>     * TC_00_Qrexec_whonix-gateway-17: [test_098_qrexec_service_socket_vm_eof](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_whonix-gateway-17/23) (failure)
>       `AssertionError: service timeout, probably EOF wasn't transferred to...`
> 
>     * TC_00_Qrexec_whonix-workstation-17: [test_053_qrexec_vm_service_eof_reverse](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_whonix-workstation-17/4) (failure)
>       `AssertionError: Timeout, probably EOF wasn't transferred`
> 
>     * TC_00_Qrexec_whonix-workstation-17: [test_092_qrexec_service_socket_dom0_eof_reverse](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_whonix-workstation-17/18) (failure)
>       `AssertionError: service timeout, probably EOF wasn't transferred fr...`
> 
>     * TC_00_Qrexec_whonix-workstation-17: [test_098_qrexec_service_socket_vm_eof](https://openqa.qubes-os.org/tests/125357#step/TC_00_Qrexec_whonix-workstation-17/23) (failure)
>       `AssertionError: service timeout, probably EOF wasn't transferred to...`

This is the only qrexec PR in this test run, so the above failures seems to be regression caused by this change.

libqrexec/libqrexec-utils.h
```diff
@@ -162,6 +162,11 @@ void buffer_append(struct buffer *b, const char *data, int len);
 void buffer_remove(struct buffer *b, int len);
 int buffer_len(struct buffer *b);
 void *buffer_data(struct buffer *b);
+/* Open /dev/null and keep it from being closed before the exec func is called.
```
marmarek, line 165: Isn't it simpler (and safer) to simply open /dev/null just before dup-ing it over 0,1,2 (in the child process already)?

marmarek:
But also, I question usefulness of this PR as a whole - the closing of standard FDs happens in a process that has a single purpose - wait for the child process and then exit, in the very same function as closing happens. There are few PAM cleanup calls, but it's very unlikely for them to be a problem (especially, it isn't a problem now, or for the last 10 or so years).

DemiMarie:
Whether or not the last commit in the PR is merged, I definitely think the _other_ commits should be merged.  In particular, it turned out that the “close the FD” functionality had no unit tests because the unit tests took a codepath that was too different from the production code.  This PR makes the production and test code follow the same path, with the result that the actual bug (`/dev/null` FD being closed by `fix_fds()`) is now caught.  I think that this test improvement (and the other bug fixes) is itself useful.

> There are few PAM cleanup calls, but it's very unlikely for them to be a problem (especially, it isn't a problem now, or for the last 10 or so years).

PAM cleanup calls into PAM modules, so it can do _anything_.  I suspect Qubes OS only gets away with it because we have a fairly simple PAM stack by default.  PAM cleanup is used for e.g. unmounting filesystems and closing encrypted volumes.

The best approach would be for PAM to run with stdin pointed at `/dev/null` and stdout and stderr pointed at the system log.  The FDs would be fixed directly before executing the child process.  That’s a bigger refactor, though.

marmarek:
> Whether or not the last commit in the PR is merged,

Yes, I was talking about the last commit (which until the last force-push was the only commit in this PR).

> PAM cleanup calls into PAM modules, so it can do _anything_. I suspect Qubes OS only gets away with it because we have a fairly simple PAM stack by default.

Aren't PAM modules expected handle proper logging themselves? I don't think they are supposed to touch calling process's stdin/out/err in any case. And if they would do, that likely would interfere also with cases where they aren't closed (and then replaced with with unrelated thing) - for example it could interfere with an application log file on stderr that is expected in a specific format (different than PAM messages).

marmarek:
As for the other commits, won't that have some non-trivial conflicts with #141 (which I hope is quite close to merge-able state)?

DemiMarie:
> As for the other commits, won't that have some non-trivial conflicts with #141 (which I hope is quite close to merge-able state)?

I can include them in #141 or rebase this PR on top of it.  I can also close this PR if you prefer, but I’d prefer that at least the bug fixes and testability changes go in.

libqrexec/libqrexec-utils.h
DemiMarie: It’s definitely simpler, but I didn’t want the extra call to `open()`.  That said, giving each child process the same open file description is less than great, as it introduces shared state that should not be there.  I don’t know if `/dev/null` has any such state, but still it isn’t great, so I went ahead and switched to your approach.

marmarek (dismissed):

marmarek:
PipelineRetry

marmarek:
One of the tests fails, I'll retry just in case, but it might be real bug

DemiMarie:
Bug is real, but this code is just the messenger: #141 uses a shell script to run the command in the no-fork-server case but not otherwise.  However, `/bin/sh` will fall back to interpreting the file as shell commands if `execve()` fails with `-ENOEXEC` and the file is not a binary file.

The proper solution is to write an extremely trivial C program (`qrexec-exec-qubes-rpc`?) that wraps `exec_qubes_rpc2()` and drop the last argument from that function.  For this PR, though, I’ll just fixup the test to work around the issue.

marmarek:
> Bug is real, but this code is just the messenger: #141 uses a shell script to run the command in the no-fork-server case but not otherwise. However, `/bin/sh` will fall back to interpreting the file as shell commands if `execve()` fails with `-ENOEXEC` and the file is not a binary file.
> 
> The proper solution is to write an extremely trivial C program (`qrexec-exec-qubes-rpc`?) that wraps `exec_qubes_rpc2()` and drop the last argument from that function. For this PR, though, I’ll just fixup the test to work around the issue.

The reason for using shell there was to load relevant environment variables (wrong PATH is what we noticed, but surely it applies to many others too, including various XDG_, DBUS_, locale etc). So, just dropping the shell will re-introduce the issue.

agent/qrexec-agent.c
```diff
@@ -120,6 +126,86 @@ static struct pam_conv conv = {
     NULL
 };
 #endif
+
+_Noreturn static void really_exec(const char *prog, const struct passwd *pw,
+                                  char **env, const char *cmd, pid_t agent_pid)
+{
+    /* 1 means init, which is completely wrong. */
+    if (agent_pid < 2) {
```
marmarek, line 134: Why is it wrong? I can imagine legitimate case of using qrexec as init (maybe some version of a stubdomain, or some other very specialized domain?

agent/qrexec-agent.c
```diff
@@ -120,6 +126,86 @@ static struct pam_conv conv = {
...
+    /* call QUBESRPC if requested */
+    if (prog) {
+        /* Set up environment variables for a login shell. */
+        char *pid_str;
+        if (asprintf(&pid_str, "%ju", (uintmax_t)agent_pid) < 1) {
+            PERROR("asprintf");
+            _exit(QREXEC_EXIT_PROBLEM);
+        }
+        const char *buf[] = {"sh", "-lc", "exec \"$@\"", "sh", exec_program, pid_str, prog, cmd, service_path, NULL};
+        /* TODO: use the user's shell not /bin/sh */
```
marmarek, line 157: While it may sounds like a smart move, it's also problematic, as the script might need to be adjusted for such shell. AFAIR "user shell" doesn't even need to be POSIX compliant.

marmarek:
> So, just dropping the shell will re-introduce the issue.

Ah, I see what you mean. Ok, I guess that should work.

marmarek:
black complains

DemiMarie:
> A lot of tests has failed, see openqa report from the bot

I’ll work on fixing this.

DemiMarie:
I did some tests locally and it looks like the existing code already has a problem (one of the tests fails intermittently when I run it in a loop).

marmarek:
Maybe there are some other problems too, but clearly several tests reliably fail with this PR, while they reliably pass without it. Do you want access to a VM with this PR installed?

DemiMarie:
> Maybe there are some other problems too, but clearly several tests reliably fail with this PR, while they reliably pass without it. Do you want access to a VM with this PR installed?

I already have one, and I can indeed reproduce the bug.  It appears that at least some tests fail _more often_ with this PR than without it, but they fail _occasionally_ even without the PR, as shown by a `while sudo qubes.tests.run TEST; do :; done` loop terminating fairly quickly.  This makes me suspect that there is a pre-existing bug (probably a race condition) in either qrexec or the tests, and this code happens to make the race condition more likely to trigger.

The way I found this is that I tried to bisect the bug.  I found 9b2e8fceee7034bb3332907650da5b59e32c7fc3 (v4.3.3 + bug fixes) was bad if I ran one test in a loop, which in turn led me to try 63e069907918f7dcffad558bbe0ab2177ee471da (v4.3.3 itself) which also failed.

marmarek:
> * system_tests_qrexec
>   
>   * TC_00_Qrexec_fedora-41-xfce: [test_080_qrexec_service_argument_allow_default](https://openqa.qubes-os.org/tests/128904#step/TC_00_Qrexec_fedora-41-xfce/11) (error)
>     `subprocess.CalledProcessError: Command '/usr/lib/qubes/qrexec-clien...`
> 
> *

Recent run with https://github.com/QubesOS/qubes-core-qrexec/commit/db5faa995d40ea3df61d0ddf5bc2e605cf8b26e1 has only one failure (1 failed out of 4 times this test was run)

I tried to get more logs, but it isn't very interesting:
```
2025-02-16 07:58:16.551 qrexec-fork-server[1054]: qrexec-agent-data.c:274:handle_new_process_common: executed: QUBESRPC test.Argument+argument test-inst-vm1 (pid 1056)
2025-02-16 07:58:16.553 qrexec-fork-server[1054]: qrexec-agent-data.c:309:handle_new_process_common: pid 1056 exited with 125
```
I don't see actual reason for the failure here...

marmarek:
This PR (currently) introduces a regression, and since it isn't very useful anyway (https://github.com/QubesOS/qubes-core-qrexec/pull/186#issuecomment-2585488504) and Demi left the team, lets abandon it instead of handing it to somebody else.

PR #191, opened 2025-02-17
refactor: drop gbulb dependency

CertainLach:
gbulb package is deprecated by upstream:
https://github.com/beeware/gbulb?tab=readme-ov-file#this-project-has-been-archived

Not sure if that is the correct way to do things, but the new implementation works on NixOS-qubes.
It seems like gobject version in qubesos is recent enough too.

### DIFF
--- a/ci/requirements.txt
+++ b/ci/requirements.txt
@@ -6,4 +6,3 @@ sphinx
 pytest-asyncio
 pyinotify
 psutil
-gbulb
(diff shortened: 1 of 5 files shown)


marmarek:
If the implementation in glib is indeed mature enough, I'll be happy to use it. Do you know which version is required? I'm looking if this won't break Debian...

CertainLach:
Oh, I haven't looked at changelog carefully enough, it is only available since 3.50 (I thought it was 3.42).

Ok, seems like it should be postponed for QubesOS, I'll maintain this patch on my own for NixOS for now.

PR #194, opened 2025-03-19
Define additional default RPC config paths

ArrayBolt3:
Originally, libqrexec only scanned /etc/qubes/rpc-config for RPC configuration. This prevented having truly ephemeral RPC config under /run/qubes/rpc-config, and also didn't allow local configuration to be stored under /usr/local/etc/qubes/rpc-config. Add both of these paths to QUBES_RPC_CONFIG_PATH.

/run/qubes/rpc-config needs some SELinux configuration for qrexec components to be able to access it, so add the needed config for that too.

Fixes https://github.com/QubesOS/qubes-issues/issues/9824

(Note, I'm not sure I did the `/run/qubes/rpc-config` directory creation correctly - I considered using `mkdir --parents` there, but was afraid of messing up the SELinux labeling on `/run/qubes` on accident by doing that. Instead, I opted to just make `/run/qubes/rpc-config`, allow the command to fail if a parent directory didn't exist, and also set the SELinux type in the `mkdir` command rather than using `restorecon` to set it right since I felt like `restorecon` was too hacky for this. If there's a better way of doing this, I'm more than happy to change it.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,8 @@ install-base: all-base
 	install -d $(DESTDIR)/etc/xdg/autostart -m 755
 	install -m 644 policy-agent-extra/qrexec-policy-agent.desktop \
 		$(DESTDIR)/etc/xdg/autostart/qrexec-policy-agent.desktop
+	install -m 755 -d $(DESTDIR)/usr/lib/tmpfiles.d
+	install -m 644 -t $(DESTDIR)/usr/lib/tmpfiles.d/ systemd/qrexec-volatile-config.conf
 .PHONY: install-base
 
 
--- a/debian/qubes-core-qrexec.install
+++ b/debian/qubes-core-qrexec.install
@@ -22,4 +22,5 @@ usr/lib/qubes/qrexec-client-vm
 usr/lib/qubes/qrexec_client_vm
 usr/lib/qubes/qubes-rpc-multiplexer
 usr/lib/qubes/qrexec-policy-agent-autostart
+usr/lib/tmpfiles.d/qrexec-volatile-config.conf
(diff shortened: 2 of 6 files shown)


systemd/qubes-qrexec-agent.service
```diff
@@ -5,6 +5,7 @@ After=xendriverdomain.service systemd-user-sessions.service
 [Service]
 Type=notify
 ExecStartPre=+/bin/sh -c '[ -e /dev/xen/evtchn ] || modprobe xen_evtchn'
+ExecStartPre=-+/bin/sh -c '[ ! -d /run/qubes/rpc-config ] && mkdir -Z etc_t /run/qubes/rpc-config'
```
marmarek, line 8: Is that -Z really needed? Or is that mkdir even appropriate here? Generally, SELinux should set correct labels when you create the directory yourself (with the above change in .fc file). And if pre-creating the directory is necessary anyway, better do that via tmpfiles mechanism.

ArrayBolt3: At least when I try making the directory manually in a shell, it ends up with the type `qubes_var_run_t`, despite the change. I can run `sudo restorecon /run/qubes/rpc-config` to fix it, but that's hacky.

Will try the systemd-tmpfiles mechanism to see if it works better.

ArrayBolt3: Implemented, tested, works.

ArrayBolt3:
Added a bit of code for allowing loading services from /run/qubes-rpc too.

marmarek (approved):

PR #198, opened 2025-04-24
Remove internal qubes from being target of ask

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/1512


marmarek:
@deeplow do you have any opinion on this regarding your usage of "internal" qubes? With this change it's still possible to make qrexec calls to them if the policy says "allow", not cannot be targeted with "ask" anymore.

marmarek:
Ok, thanks :)

@ben-grande now just black complains

ben-grande:
Resolved merge conflicts.

ben-grande:
> https://openqa.qubes-os.org/tests/139759#step/TC_20_DispVM_debian-12-xfce/6
> https://openqa.qubes-os.org/tests/139759#step/TC_20_DispVM_fedora-41-xfce/6

Something not being cleaned and them every dispvm test after, broken because of that.

marmarek:
> Something not being cleaned and them every dispvm test after, broken because of that.

Yes, some object leaked. We have quite strict checking for that between tests. Outside of tests it's easy to miss - usually it will "just" result in a memory leak that will accumulate over time so you'll notice it only after days/weeks of uptime. The test VM (intentionally) has just 8GB of memory, so issues like this are noticed earlier, even if the leak check missed it. And indeed, it looks like it happened here too:
```
ERROR:vm.disp8059:Start failed: Not enough memory to start domain 'disp8059'
VM disp8059 start failed at 2025-05-21 21:39:51
WARNING:vm.test-inst-dvm:Not preloading '1' disposable(s) due to insufficient memory
ERROR
```

See also terminal log from the test run: https://openqa.qubes-os.org/tests/139759/file/system_tests-tests-qubes.tests.integ.dispvm.log looks like preloading remains running beyond test end, you may need to extend `tearDown` of that test class (possibly to interrupt preloading? maybe killing all dispvms being preloaded will be enough?). The `AttributeError: 'DispVM' object has no attribute 'app'` may look confusing, but it happens because test cleanup already kicked in, while some background job (preloading?) was attempting to access VM objects.

Quite possibly you may need to save a reference to tasks scheduled in background (for cases of ensure_future instead of await), at least to wait for them in the tearDown method.

ben-grande:
> you may need to extend `tearDown`

Makes sense, preloading by setting feature is a future, so I will make sure to wait for every domain that was preloaded to  not be in the vm collection anymore.

> See also terminal log from the test run: https://openqa.qubes-os.org/tests/139759/file/system_tests-tests-qubes.tests.integ.dispvm.log

https://openqa.qubes-os.org/tests/139759/#downloads

And now I see I can download the tests logs, that is useful.

>  Failed to create preloaded disposable, limit of preloaded DispVMs reached

Just note that although this is an exception and clutters the log, it happens because the `used` event is being simultaneously called 5 times and each of those 5 try to preload 5 new ones. This happens because I changed the `used` event to also refill preloaded disposables, but that doesn't work well if multiple are used at the same time. It does not cause anything to fail and preloads all that is necessary, the exception is ignored because the `used` event is called as `future`, only problem is the log garbage.

I could use an async lock but that would make it preload slower (think of 1 qube being used and them 4 qubes being used, waiting for 1 to become preloaded to start preloading other 4 seems inefficient) or I could handle the exception with a `pass` as the code above checks how many it could preload.

What do you think?

marmarek:
> Just note that although this is an exception and clutters the log, it happens because the `used` event is being simultaneously called 5 times and each of those 5 try to preload 5 new ones.

Maybe change that from exception to a warning log message (and simple return)?

ben-grande:
> Maybe change that from exception to a warning log message (and simple return)?

Also possible and I am taking this route then.

marmarek (approved):

PR #200, opened 2025-04-30
qrexec-policy-agent: Replace usage of deprecated Gtk.Entry.set_icon_from_stock

HW42:
Gtk docs suggest set_icon_from_icon_name as replacement. But for some reason that uses another theme. So instead reuse the code from the fallback in _get_icon.


marmarek:
pylint and black complain

marmarek (approved):

PR #204, opened 2025-05-07
Don't translate dom0 to @adminvm

HW42:
The previous behavior of handling "dom0" and "@adminvm" in policies was
inconsistent and let to the problem that you could not specify
target=dom0 default_target=dom0 (or both @adminvm) since one was
translated to @adminvm and the other was alsways expanded to dom0.

Based on the discussion in https://github.com/QubesOS/qubes-issues/issues/9870 we now stop to
translate dom0 (or it's UUID) to @adminvm. The later works as keyword
that is expanded to dom0.

Note this also uncovered some wrong tests (like that "@type:AdminVM"
should not match "dom0").

Closes https://github.com/QubesOS/qubes-issues/issues/9870

---

Depends on #203

### DIFF
--- a/qrexec/policy/parser.py
+++ b/qrexec/policy/parser.py
@@ -236,6 +236,14 @@ def match_strings(info: SystemInfo, self: str, other: str) -> bool:
     return self == other
 
 
+def is_dom0(token) -> bool:
+    return token in (
+        "@adminvm",
+        "dom0",
+        "uuid:00000000-0000-0000-0000-000000000000",
+    )
+
+
 class VMToken(str, metaclass=VMTokenMeta):
     """A domain specification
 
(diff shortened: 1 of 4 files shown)


marmarek:
Indeed there are problematic cases. With sys-gui, clipboard paste doesn't work:
```
May 08 18:07:08.218152 dom0 qrexec-policy-daemon[1725]: qrexec: policy.EvalGUI+qubes.ClipboardPaste: sys-gui -> @adminvm: denied: no matching rule found
```

The policy is here: https://github.com/QubesOS/qubes-mgmt-salt-dom0-virtual-machines/blob/main/qvm/template-gui.jinja#L15 and it has `dom0` as a target.

marmarek:
I must admit it's a weird case. But it worked before, and we have no idea where is is that used (in various user scripts etc). So, I guess it needs to remain working.
BTW, with RemoteVM, there may be a point where `@adminvm` might get smarter and point not at local dom0, but dom0 of a given VM (so Admin API would work on RemoteVM too). But R4.3 is not the time for that, there are a lot of details that aren't clear yet (including whether it's a good idea at all). But it would be consistent with translating `@adminvm` to `dom0` right now.

marmarek:
PipelineRetry

marmarek (approved):

PR #206, opened 2025-05-18
Allow individual VMs to decide their default user for actions

ArrayBolt3:
Still WIP, entirely untested (though it does build at least). Posting mostly to get preliminary feedback while I work on testing it.


ArrayBolt3:
This probably isn't useful after all, and has security concerns. Closing.

PR #207, opened 2025-05-20
Resolve merge conflict for #198

marmarek:


ben-grande:
> https://openqa.qubes-os.org/tests/139585#step/TC_20_DispVM_debian-12-xfce/4

This error is expected as it was tested against #660 before I pushed the changes to fix the integration tests.

> https://openqa.qubes-os.org/tests/139585#step/TC_20_DispVM_debian-12-xfce/6

May be due to not having enough memory to preload that many qubes and use them, overload of having 5 preloaded qubes plus preloading 5 as each of them is used. Will wait for the newest version of #660 to be tested to be sure.

marmarek:
> > qubes.exc.QubesValueError: Qube GUI is 'False' and does not support the RPC 'qubes.WaitForRunningSystem'
> 
> Missing [QubesOS/qubes-core-agent-linux#571](https://github.com/QubesOS/qubes-core-agent-linux/pull/571) installed in the templates.

Actually, this is more likely update issue that didn't refreshed supported rpc list.

PR #213, opened 2025-08-06
Implement custom RPC confirmation prompt

HW42:
For now only qubes.AuthorizeInVMAdminAccess is customized.

QubesOS/qubes-issues#5853


marmarek:
rpm fails to build

qrexec/rpcconfirmation/admin_authz.py
```diff
@@ -0,0 +1,61 @@
...
+        os.path.join("glade", "AdminAuthzRPCConfirmationWindow.glade")
+    )
+
+    @staticmethod
+    def match(entries_info, source, service, argument, targets_list, target):
+        # pylint: disable=too-many-positional-arguments
+        return (
+            service == "qubes.AuthorizeAdminAccess"
+            and targets_list == ["dom0"]
+            and target == "dom0"
```
marmarek, line 35: I wonder if matching on `target==dom0` is really necessary, seems to not change much in terms of usefulness of this custom confirmation? OTOH `targets_list==["dom0"]` very much makes sense.

qrexec/tools/qrexec_policy_agent.py
```diff
@@ -46,534 +41,36 @@
...
 ):
-    # pylint: disable=too-many-arguments,too-many-positional-arguments
-    window = RPCConfirmationWindow(
-        entries_info, source, service, argument, targets_list, target
-    )
+    # pylint: disable=too-many-positional-arguments
+    for win_class in [
+        AdminAuthzRPCConfirmationWindow,
+        RPCConfirmationWindow,  # matches every request
+    ]:
```
marmarek, line 62: There are probably smarter (easier to extend) methods of doing this, but as discussed before, lets not worry about it right now.

marmarek (changes requested):
The code structure is okay, and also it seems to work just fine :)

But, please split code movement (extracting the generic dialog to a separate file) into a separate commit. And then add the first custom prompt on top of it.

PR #214, opened 2025-08-20
Minimal update of the documentation

parulin:
This is a minimal update, in order to get a less outdated documentation.

That is still far from perfect...


marmarek:
All packages failed to build - you need to replace recommonmark with myst-parser in package (both rpm and deb) build dependencies too.

marmarek (changes requested):
see above

parulin:
Ok, sorry for taking your time. Changing a doc dependency wasn't as simple as I expected...

I added `python3-myst-parser` to the build dependencies of debian even if recommonmark wasn't there.

marmarek (approved):

PR #222, opened 2026-02-28
qrexec-policy-graph: add qube colors

### LINKED ISSUE #3007: qrexec-policy-graph: color nodes according to actual colors

notmuchtohide:
Fixes QubesOS/qubes-issues#3007. I have not fixed the tests yet, but I believe the rest is ready for feedback.

Code decisions:
  1. Node styling is not organized: Node styling is added as nodes are found instead of being grouped together. It should have no impact on the final graph and was considered as a potentially acceptable trade-off for implementation simplicity.
  2. qubesadmin vs. system_info: The already used system_info did not provide qube colors, therefore it was obtained from qubesadmin.

Styling decisions:
  1. Qube color was added to node borders in case of a literal qube (one that doesn't start with "@"). On non-literal qubes, nodes are dotted because it's not necessarily an individual qube.
  2. The chosen style of coloring borders instead of solid filling made it easier to deal with the inner text (example: black text on black colored qubes).

## Before
before 

## After
after

### DIFF
--- a/qrexec/tests/tools/qrexec_legacy_convert.py
+++ b/qrexec/tests/tools/qrexec_legacy_convert.py
@@ -31,6 +31,7 @@ def system_info():
     system_info = {
         "dom0": {
             "icon": "black",
+            "label": "0xffffff",
             "template_for_dispvms": False,
             "guivm": None,
             "type": "AdminVM",
@@ -39,6 +40,7 @@ def system_info():
         },
         "work": {
             "icon": "red",
+            "label": "0xcc0000",
             "template_for_dispvms": False,
             "guivm": None,
             "type": "AppVM",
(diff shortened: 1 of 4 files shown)


marmarek:
> [CI failure](https://gitlab.com/QubesOS/qubes-core-qrexec/-/jobs/13408625575) seems to be related to missing the PR on core-admin.

Tests have mockup system info structure, see qrexec/tests/policy_graph.py.  And I'm pretty sure some of the tests will need an update to include the label too.

ben-grande:
Passed CI: https://gitlab.com/QubesOS/qubes-core-qrexec/-/pipelines/2373959020

gi problem for pylint, will ignore that and let this PR be scheduled for openqa.

marmarek (approved):

PR #223, opened 2026-03-02
Improve policy admin client and related tools

### LINKED ISSUE #10745: qubes-policy-editor (CLI) exits prematurely on failure to replace the policy, losing changes
### Qubes OS release

Qubes OS 4.3

### Brief summary

Attempting to access the same policy, making changes on one version, saving it and going to the second window and trying to save fails. It is nice to see the error, but it quits the program, which means the user will lose the changes he made. Of course he can't save easily because of the current token, but it is better to let the user save a backup of the policy in this case, which means, it shouldn't exit prematurely.

### Steps to reproduce

1. `qubes-policy-editor`
2. On another prompt, `qubes-policy-editor`
3. On one of the editors, modify (extra line), save and close)
4. On the other editor, try to save

### Expected behavior

Shows the error and save the file, printing the file name. This allows for the user to copy the contents from that file when reopening the real policy.

### Actual behavior

```python3
Failed to replace file: 30-user
subprocess.CalledProcessError: Command '['/usr/bin/qrexec-client', '-d', 'dom0', 'DEFAULT:QUBESRPC policy.Replace+30-user dom0']' returned non-zero exit status 2.
```

### Additional information

The traceback doesn't help:

```python3
  File "/usr/lib/python3.13/site-packages/qrexec/tools/qubes_policy_editor.py", line 101, in manage_policy
    policy_replace(name, content, token)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qrexec/policy/admin_client.py", line 57, in policy_replace
    self.call("policy.Replace", name, token + "\n" + content)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qrexec/policy/admin_client.py", line 74, in call
    return call("dom0", service_name, arg, input=payload)
  File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 53, in call
    return subprocess.check_output(
           ~~~~~~~~~~~~~~~~~~~~~~~^
        command, **prepare_subprocess_kwds(input)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ).decode()
    ^
  File "/usr/lib64/python3.13/subprocess.py", line 472, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
               **kwargs).stdout
               ^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/usr/bin/qrexec-client', '-d', 'dom0', 'DEFAULT:QUBESRPC policy.Replace+30-user dom0']' returned non-zero exit status 2.
```

As the exceptions are not being sent to the user. I believe they should and I will open another issue about this.

### LINKED ISSUE #10746: Report qrexec.admin exception messages back to the caller
### The problem you're addressing (if any)

The exception is not sent back to the caller. Instead, it raises generic `subprocess.CalledProcessError`:

```python3
  File "/usr/lib/python3.13/site-packages/qrexec/tools/qubes_policy_editor.py", line 101, in manage_policy
    policy_replace(name, content, token)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qrexec/policy/admin_client.py", line 57, in policy_replace
    self.call("policy.Replace", name, token + "\n" + content)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qrexec/policy/admin_client.py", line 74, in call
    return call("dom0", service_name, arg, input=payload)
  File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 53, in call
    return subprocess.check_output(
           ~~~~~~~~~~~~~~~~~~~~~~~^
        command, **prepare_subprocess_kwds(input)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ).decode()
    ^
  File "/usr/lib64/python3.13/subprocess.py", line 472, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
               **kwargs).stdout
               ^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/usr/bin/qrexec-client', '-d', 'dom0', 'DEFAULT:QUBESRPC policy.Replace+30-user dom0']' returned non-zero exit status 2.
```
https://github.com/QubesOS/qubes-issues/issues/10745

Which is pretty bad to debug. And some tools recommend you to check `/var/log/qubes/policy-admin.log` on `dom0` (`tools/qubes_policy_admin.py`), which the GUI uses somehow.

### The solution you'd like

Because it is using the shell to call `qrexec-client`, the client doesn't know exceptions, it needs to be a message sent to stderr. I will check if there is any private information that shouldn't be sent in the exceptions before making such a change.

I think that both `PolicyAdminException` and `PolicyAdminTokenException` should send back the messages to the client.

### The value to a user and who that user might be

When an error occurs, user will know what the error is.
If the developer needs to debug the error a user shared, it is better if it is informative.

### Completion criteria checklist

- [ ] pr to core-qrexec
- [ ] pr to desktop-linux-manager

ben-grande:
- reduce number of statements
- pass `--` to shell commands
- prints traeback on errors
- custom `.policy` extension so editors can recognize the filetype and load syntax accordingly
- remove the non-root blocker, unnecessary since using PolicyClient 

And many more things as commit grew. Check each commit message.

Fixes: https://github.com/qubesos/qubes-issues/issues/10745
Fixes: https://github.com/qubesos/qubes-issues/issues/10746


marmarek (changes requested):

ben-grande:
> /usr/bin/xvfb-run: line 186: kill: (2943) - No such process

Attempting to address this in https://github.com/QubesOS/qubes-core-qrexec/pull/228.

marmarek:
Ok, so now is time to fix all the fallout of changed public API (especially: changed exceptions). First victim: qubes-global-config:
```
  File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 66, in call
    return subprocess.check_output(command, **kwds).decode()
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 472, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
               **kwargs).stdout
               ^^^^^^^^^
  File "/usr/lib64/python3.13/subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/usr/bin/qrexec-client', '-d', 'dom0', 'DEFAULT:QUBESRPC policy.Get+50-config-u2f dom0']' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

  File "/usr/lib/python3.13/site-packages/qubes_config/global_config/global_config.py", line 304, in do_activate
    self.perform_setup()
    ~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/qubes_config/global_config/global_config.py", line 414, in perform_setup
    self.handlers["usb"] = DevicesHandler(
                           ~~~~~~~~~~~~~~^
        self.qapp, self.policy_manager, self.builder
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.13/site-packages/qubes_config/global_config/usb_devices.py", line 726, in __init__
    self.u2f_handler = U2FPolicyHandler(
                       ~~~~~~~~~~~~~~~~^
        self.qapp, self.policy_manager, gtk_builder, usb_qubes
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.13/site-packages/qubes_config/global_config/usb_devices.py", line 312, in __init__
    self._initialize_data()
    ~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/qubes_config/global_config/usb_devices.py", line 394, in _initialize_data
    self.rules, self.current_token = self.policy_manager.get_rules_from_filename(
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        self.policy_filename, self.default_policy
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.13/site-packages/qubes_config/global_config/policy_manager.py", line 85, in get_rules_from_filename
    rules_text, token = self.policy_client.policy_get(filename)
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/qrexec/policy/admin_client.py", line 59, in policy_get
    token, content = self.call(
                     ~~~~~~~~~^
        rpc=rpc, arg=policy, is_include=is_include
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ).split("\n", 1)
    ^
  File "/usr/lib/python3.13/site-packages/qrexec/policy/admin_client.py", line 121, in call
    return call(dest="dom0", rpcname=rpc, arg=arg, payload=payload)
  File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 74, in call
    raise exception(stderr_exc_msg) from exc
qrexec.policy.admin.PolicyAdminFileNotFoundException: Not found: /etc/qubes/policy.d/50-config-u2f.policy
```

The whole `PolicyManager` class there expects plain `CalledProcessError` exception...
It would be good to make it work both with old and new code - including old backend side (so it can run on 4.3 too, not only on "devel" aka 4.4+). Which means accepting both old and new exceptions there.
And to make it work with older qrexec, you can do something like:
```python
try:
    from qrexec.admin import PolicyAdminFileNotFoundException
except ImportError:
    PolicyAdminFileNotFoundException = subprocess.CalledProcessError
```
(this does not replace accepting `CalledProcessError` too in relevant `except` clauses)

marmarek:
Alternatively, make the new exception inherit _also_ from `CalledProcessError`, then no other changes should be needed (hopefully).

ben-grande:
Somehow, the bot didn't post a comment on this PR. Only relevant test seems to be [system_tests_gui_tools](https://openqa.qubes-os.org/tests/172479), with the [.xsession-errors.log](https://openqa.qubes-os.org/tests/172479/file/desktop_linux_manager_config-journalctl.log).

> ```python
> try:
>     from qrexec.admin import PolicyAdminFileNotFoundException
> except ImportError:
>     PolicyAdminFileNotFoundException = subprocess.CalledProcessError
> ```

I am more inclined to use this instead of inheriting from `subprocess.CalledProcessError` because the latter is too broad and I didn't like that it was guessing that a subprocess exception means that the file is not found, and not anything else that might have failed.

ben-grande:
I will use existing PR: https://github.com/QubesOS/qubes-desktop-linux-manager/pull/301

ben-grande:
openQArun PR_LABEL=openqa-pending TEST=system_tests_gui_tools

ben-grande:
Marek, do you know why the command isn't triggering the run?

marmarek:
This repo is not on the list: https://github.com/QubesOS/openqa-tests-qubesos/blob/main/api/gitlab_api.py#L31

Anyway, I started it now: https://gitlab.com/QubesOS/qubes-continuous-integration/-/pipelines/2438584386

ben-grande:
> This repo is not on the list: https://github.com/QubesOS/openqa-tests-qubesos/blob/main/api/gitlab_api.py#L31

I always forget to check this. Last time you said there was a reason to not allow more repos, which I do not remember.

marmarek:
For the command to work (without explicit label), `.gitlab-ci.yml` needs to include `gitlab-host-vm-openqa.yml` (or `gitlab-host-openqa.yml`) config. That doesn't apply to command with LABEL=..., but then it also doesn't matter in which PR such command is issued.

qrexec/client.py
marmarek, line 1: Uh, sorry for not noticing before, but changes in this file are very wrong... `qrexec.client` module is about making generic qrexec calls from python, not just for managing policies. It must _not_ expect any specific format of stderr. Policy admin methods client should be limited to admin_client.py.
And also, it's a public API, don't change it (speaking about keyword parameters rename you did).

marmarek (changes requested):

qrexec/client.py
ben-grande: > `qrexec.client` module is about making generic qrexec calls from python, not just for managing policies. It must _not_ expect any specific format of stderr.

It checks if the first item of stderr is the same as an exception class. It is not idiomatic, but the call is made via a subprocess/`qrexec-client-vm`, so, uh... do you have any ideas?

How the admin-client does it:

- Get stdout: https://github.com/QubesOS/qubes-core-admin-client/blob/38c79ef29d41b7aca08cfeb8b2b5e86f3f3d63fe/qubesadmin/app.py#L1078
- Parse stdout to predefined format null separated and get exception: https://github.com/QubesOS/qubes-core-admin-client/blob/38c79ef29d41b7aca08cfeb8b2b5e86f3f3d63fe/qubesadmin/base.py#L134

8d5787f23f0468b4032855b60f0acbc220ffc1e2: "Return policy admin exceptions to client stderr". This is writing the exception to stderr. The [stdout is already being used](https://github.com/QubesOS/qubes-core-qrexec/blob/9082bbceb6992479dfa68e54777f8eb63b839172/qrexec/tools/qubes_policy_admin.py#L71) for sending the data. I don't think we can easily make it null delimited to support more fields and exceptions without making it a breaking change. Do you have any recommendations on what to do in this case?

> Policy admin methods client should be limited to admin_client.py.

The change was just to raise the exception. I can make it only on call method of `admin_client.py`, after we solve the issue of the point above.

> And also, it's a public API, don't change it (speaking about keyword parameters rename you did).

I guess this is 1ba1552ade7ccc8f3a5d5402f621fd24c3761e2f, "Fix input builtin redefintion". The downside of using a variable with the same name as a builtin is that parsing might become confusing, in some places, it will use the value of the builtin if not redefined yet, for example, leading to strange conditions. I can drop that commit, maybe we can add it to the next breaking change of qrexec.

marmarek: > The [stdout is already being used](https://github.com/QubesOS/qubes-core-qrexec/blob/9082bbceb6992479dfa68e54777f8eb63b839172/qrexec/tools/qubes_policy_admin.py#L71) for sending the data. I don't think we can easily make it null delimited to support more fields and exceptions without making it a breaking change.

Indeed, I don't think this stdout can be used here.
Stderr is fine, but `qrexec.client` module shouldn't parse it. Move that parsing (catching CalledProcessError) to admin_client.py - just that.

> I guess this is [1ba1552](https://github.com/QubesOS/qubes-core-qrexec/commit/1ba1552ade7ccc8f3a5d5402f621fd24c3761e2f), "Fix input builtin redefintion".

I know `input` is (also) a builtin name, but it's here to stay. It mirrors the `subprocess` standard library, which also uses `input=` as the argument name for the same purpose.

ben-grande:
Added tests for `policy/admin_client.py`.

ben-grande:
Do you prefer max-line-length 88 or 100 for this repo? It seems that it is being blackened as 88 and sometimes 100 but pylint is checking 100.

There doesn't seem to be an open PR that will be impacted by the reformatting.

marmarek:
> Do you prefer max-line-length 88 or 100 for this repo? It seems that it is being blackened as 88 and sometimes 100 but pylint is checking 100.

It's actually 80, not 88, enforced by black.

qrexec/client.py
```diff
@@ -71,26 +70,10 @@ async def call_async(dest, rpcname, arg=None, *, input=None):
...
-        # Assume this is a file
-        stdin = input
-        to_communicate = None
-
-    process = await asyncio.create_subprocess_exec(
-        *command, stdin=stdin, stdout=subprocess.PIPE
-    )
-
+    command = make_command(dest=dest, rpcname=rpcname, arg=arg)
+    kwds = prepare_subprocess_kwds(input)
```
marmarek, line 74: This doesn't set `stdout=subprocess.PIPE` anymore, which is critical for the `stdout.decode()` below to work.

Took me quite a while to realize this causes double qrexec prompts with ~GUI domain~Interface qube at https://openqa.qubes-os.org/tests/174243
The actual exception is:
```
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[18094]: allow:personal
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]: Unhandled exception in client_connected_cb
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]: transport: <_SelectorSocketTransport closed fd=10>
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]: Traceback (most recent call last):
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:   File "/usr/lib/python3.13/site-packages/qrexec/tools/qrexec_policy_daemon.py", line 129, in handle_client_connection
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     result = await handle_request(
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:              ^^^^^^^^^^^^^^^^^^^^^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:         **args, log=log, policy_cache=policy_cache
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     )
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     ^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:   File "/usr/lib/python3.13/site-packages/qrexec/tools/qrexec_policy_exec.py", line 397, in handle_request
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     return await resolution.execute()
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:   File "/usr/lib/python3.13/site-packages/qrexec/tools/qrexec_policy_exec.py", line 105, in execute
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     ask_response = await call_socket_service(
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:         guivm, service, source_domain, params
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     )
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     ^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:   File "/usr/lib/python3.13/site-packages/qrexec/server.py", line 119, in call_socket_service_remote
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     output_data = await call_async(remote_domain, service, input=input_data)
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:   File "/usr/lib/python3.13/site-packages/qrexec/client.py", line 80, in call_async
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:     return stdout.decode()
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]:            ^^^^^^^^^^^^^
Apr 22 11:11:36.816738 dom0 qrexec-policy-daemon[1364]: AttributeError: 'NoneType' object has no attribute 'decode'
```

marmarek (changes requested):

qrexec/client.py
ben-grande: Fixed: https://openqa.qubes-os.org/tests/overview?distri=qubesos&version=4.3&build=2026042312-4.3&groupid=11

ben-grande:
I am adding more tests for coverage. So I can understand why the openqa label was removed, is something failing or it is because it requires review?

marmarek:
I want to run a test with https://github.com/QubesOS/qubes-desktop-linux-manager/pull/301 but not this one, to see if it really still works with the old version.

ben-grande:
PipelineRetryFailed

marmarek (approved):

PR #224, opened 2026-03-03
Log exit code and command line on spawn failures

ben-grande:
For: https://github.com/QubesOS/qubes-issues/issues/10740

---

Logging this information makes it easier to grep on huge logs.

### DIFF
--- a/agent/qrexec-agent-data.c
+++ b/agent/qrexec-agent-data.c
@@ -202,7 +202,8 @@ static int handle_just_exec(struct qrexec_parsed_command *cmd)
 static int handle_new_process_common(
     int type, int connect_domain, int connect_port,
     struct qrexec_parsed_command *cmd,
-    int buffer_size)
+    int buffer_size,
+    const char *cmdline_for_logging)
 {
     libvchan_t *data_vchan;
     int exit_code;
@@ -257,7 +258,7 @@ static int handle_new_process_common(
                     exit_code = 0;
             }
             if (exit_code != 0) {
-                LOG(ERROR, "failed to spawn process");
+                LOG(ERROR, "failed to spawn process (exited %d): %s", exit_code, cmdline_for_logging);
                 /* Send stdout+stderr EOF first, since the service is expected to send
                  * one before exit code in case of MSG_EXEC_CMDLINE. Ignore
                  * libvchan_send error if any, as we're going to terminate soon
@@ -271,7 +272,11 @@ static int handle_new_process_common(
                 libvchan_close(data_vchan);
                 return exit_code;
             }
-            LOG(INFO, "executed: %s (pid %d)", cmd->cmdline, pid);
+            if (pid == 0)
+                LOG(INFO, "executed: %s", cmd->cmdline);
+            else
+                LOG(INFO, "executed: %s (pid %d)", cmd->cmdline, pid);
+
             break;
         default:
             LOG(ERROR, "unknown request type: %d", type);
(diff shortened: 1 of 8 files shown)


marmarek (changes requested):

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

ben-grande:
PipelineRetryFailed

ben-grande:
Marek, some weird failure: https://gitlab.com/QubesOS/qubes-core-qrexec/-/jobs/14147342413

```python3
_______ TestAgentExecQubesRpc.test_exec_service_with_invalid_config_path _______
self = <qrexec.tests.socket.agent.TestAgentExecQubesRpc testMethod=test_exec_service_with_invalid_config_path>
        def test_exec_service_with_invalid_config_path(self):
            util.make_executable_service(
                self.tempdir,
                "rpc",
                "qubes.Service",
                """\
    #!/bin/sh
    echo "arg: $1, remote domain: $QREXEC_REMOTE_DOMAIN"
    """,
            )
            config_path = os.path.join(self.tempdir, "rpc-config")
            os.rmdir(config_path)
            os.symlink("/dev/null/doesnotexist", config_path)
>           self._test_exec_service_fail()
qrexec/tests/socket/agent.py:659: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
qrexec/tests/socket/agent.py:616: in _test_exec_service_fail
    self.assertListEqual(
E   AssertionError: Lists differ: [] != [(401, b''), (402, b''), (403, b'}\x00\x00\x00')]
E   
E   Second list contains 3 additional elements.
E   First extra element 0:
E   (401, b'')
E   
E   - []
E   + [(401, b''), (402, b''), (403, b'}\x00\x00\x00')]
----------------------------- Captured stderr call -----------------------------
2026-04-29 12:16:37.017 qrexec-agent[2184]: exec.c:437:find_file: stat/lstat(/tmp/tmppsnv8acm/rpc-config/qubes.Service+arg): Not a directory
2026-04-29 12:16:37.017 qrexec-agent[2184]: qrexec-agent.c:606:handle_server_exec_request_init: Could not load config for command gitlab-runner:QUBESRPC qubes.Service+arg domX
```

marmarek:
> ```python
> 2026-04-29 12:16:37.017 qrexec-agent[2184]: qrexec-agent.c:606:handle_server_exec_request_init: Could not load config for command gitlab-runner:QUBESRPC qubes.Service+arg domX
> ```

This is expected given the test name. But it should send final packets. The test passes on main, so I think this PR introduces the regression. Maybe it crashes before sending it? Looking at the code, one explanation could be `cmd` being NULL - that would crash it exactly before sending those final packets.

ben-grande:
> `cmd` being NULL

https://github.com/QubesOS/qubes-core-qrexec/blob/f6e44eaad81e1eb2465e017b875ae36a03e42277/agent/qrexec-agent.c#L688

Here the `cmd` is NULL, but as this caller has access to `cmdline`, I think the solution is to pass it along.

marmarek:
> Here the `cmd` is NULL, but as this caller has access to `cmdline`, I think the solution is to pass it along.

In that case (`cmd==NULL`), the command line is not very interesting, it's a `SOCKETxx` string, to match outgoing request. See `handle_server_exec_request_do()`, especially comments there. In practice it will be `SOCKET12` most of the time.

So, maybe instead of passing along not very useful string, simply handle NULL cmd in that error message with some placeholder, like `outgoing call`? That would avoid confusion by having both `cmd` and `cmdline` in some of those functions (where only `cmd` should really be used).

marmarek (approved):

PR #227, opened 2026-03-21
qrexec: improve error handling for invalid daemon connection

### LINKED ISSUE #10792: negotiate_connection_params: write daemon: Bad file descriptor
### Qubes OS release

Qubes OS 4.3

### Brief summary

Message as above found in .xsession-errors after test run

### Steps to reproduce

Not sure, probably:
1. Start a qube
2. Kill it before it fully starts
3.

### Expected behavior

Cleanly handling of destroyed qube. While qrexec connection error may happen here, the `Bad file descriptor` message should not happen - it means some earlier instruction wasn't properly handled, likely the "No such file or directory" error.

### Actual behavior

```
test-inst-appvm: Failed to send monitor layout: b'2026-03-17 22:25:24.774 qrexec-client[7302]: qrexec-daemon-common.c:263:connect_unix_socket: connect /var/run/qubes/qrexec.test-inst-appvm: No such file or directory\n2026-03-17 22:25:24.774 qrexec-client[7302]: qrexec-daemon-common.c:31:negotiate_connection_params: write daemon: Bad file descriptor\n'
```

### Additional information

https://openqa.qubes-os.org/tests/170039#downloads

therohityadav:
### Details
When `connect_unix_socket()` or `connect_unix_socket_by_id()` fails,
it returns a negative file descriptor. Previously, this value could be
passed directly to functions like `negotiate_connection_params()` and
`send_service_connect()`, which then attempt to operate on it,
leading to `EBADF` ("Bad file descriptor") errors.

This patch validates the socket descriptor at the call sites and
prevents further processing if the connection fails.

### Result
- Avoids passing invalid file descriptors to downstream functions
- Prevents misleading `EBADF` errors during connection setup

Fixes QubesOS/qubes-issues#10792

### DIFF
--- a/daemon/qrexec-client.c
+++ b/daemon/qrexec-client.c
@@ -345,6 +345,9 @@ int main(int argc, char **argv)
                                    wait_connection_end, use_uuid) ? 0 : 137;
         } else {
             s = connect_unix_socket(domname);
+            if (s < 0) {
+                goto cleanup;
+            }
             if (!negotiate_connection_params(s,
                         src_domain_id,
                         just_exec ? MSG_JUST_EXEC : MSG_EXEC_CMDLINE,
--- a/daemon/qrexec-daemon-common.c
+++ b/daemon/qrexec-daemon-common.c
@@ -206,8 +206,13 @@ bool qrexec_execute_vm(const char *target, bool autostart, int remote_domain_id,
     }
 
     s = connect_unix_socket_by_id((unsigned)remote_domain_id);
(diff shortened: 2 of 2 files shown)


marmarek (changes requested):
This is _not_ correct solution. The proper solution is finding where that wrong FD is coming from and handling it there. It might be closed at some earlier point, or being invalid from the very beginning (wherever it got initialized - for example if error handling is missing there).

marmarek (approved):

PR #232, opened 2026-05-28
Log calls made from dom0 to the journal

ben-grande:
Follow dom0 calls:

        sudo journalctl -f SYSLOG_IDENTIFER=qrexec-client

### DIFF
--- a/daemon/qrexec-client.c
+++ b/daemon/qrexec-client.c
@@ -37,6 +37,7 @@
 #include <fcntl.h>
 #include <err.h>
 #include <time.h>
+#include <syslog.h>
 
 #include "libqrexec-utils.h"
 #include "qrexec-daemon-common.h"
@@ -272,6 +273,16 @@ int main(int argc, char **argv)
         usage(argv[0], 1);
     }
 
+    if (access("/run/qubes-service/qrexec-debug", F_OK) == 0) {
+        char * src_domain_for_log;
+        if (src_domain_name == NULL)
+            src_domain_for_log = "dom0";
+        else
+            src_domain_for_log = src_domain_name;
+        openlog("qrexec-client", LOG_PID, LOG_USER);
+        syslog(LOG_INFO, "qrexec: %s: %s -> %s", remote_cmdline, src_domain_for_log, domname);
+        closelog();
+    }
     if (target_refers_to_dom0(domname)) {
         if (request_id != NULL) {
             if (request_id[0] == '\0') {
--- a/run-tests
+++ b/run-tests
@@ -1,8 +1,18 @@
 #!/bin/bash -ex
 
 case $0 in (/*) cd "${0%/*}/";; (*/*) cd "./${0%/*}";; esac
+
+install_rpm_deps(){
+    set -- python3dist\({coverage,pytest,gbulb,pyinotify,pytest-asyncio}\) pam-devel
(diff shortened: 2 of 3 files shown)


daemon/qrexec-client.c
```diff
@@ -272,6 +273,13 @@ int main(int argc, char **argv)
         usage(argv[0], 1);
     }
 
+    char * src_domain_for_log;
+    if (src_domain_name == NULL)
+        src_domain_for_log = "dom0";
+    else
+        src_domain_for_log = src_domain_name;
+    sd_journal_send("MESSAGE=qrexec: %s: %s -> %s", remote_cmdline, src_domain_for_log, domname,
```
ben-grande: The format differs from `qrexec.parser` qubes-qrexec-policy-daemon.

- It uses `service+arg: src -> dest`
- Here we use `src -> dest: remote_cmdline`

Because I the cmdline may be long, I think it's best if it's at the end. But I understand this causes a bit of inconsistency when reading and comparing against other formats.

ben-grande:
https://gitlab.com/QubesOS/qubes-core-qrexec/-/jobs/14597578686

```
Package libsystemd was not found in the pkg-config search path.
Perhaps you should add the directory containing `libsystemd.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libsystemd' not found
```

```
% fd libsystemd.pc /usr
/usr/lib64/pkgconfig/libsystemd.pc

% rpm -qf /usr/lib64/pkgconfig/libsystemd.pc
systemd-devel-256.17-1.fc41.x86_64
```

---

I see it now, package is not really being installed.
```
+ sudo dnf install 'python3dist(coverage)' 'python3dist(pytest)' 'python3dist(gbulb)' 'python3dist(pyinotify)' 'python3dist(pytest-asyncio)' qubes-libvchan-xen-devel pam-devel systemd-devel
Updating and loading repositories:
Repositories loaded.
Package "python3-coverage-7.3.2-6.fc42.x86_64" is already installed.
Package "python3-pytest-8.3.4-2.fc42.noarch" is already installed.
Package "python3-inotify-0.9.6-37.fc42.noarch" is already installed.
Package "python3-pytest-asyncio-0.24.0-2.fc42.noarch" is already installed.
Package "pam-devel-1.7.0-7.fc42.x86_64" is already installed.
Package                   Arch   Version       Repository                 Size
Installing:
 python3-gbulb            noarch 0.6.5-2.fc42  fedora                219.8 KiB
 qubes-libvchan-xen-devel x86_64 4.2.7-1.fc42  qubes-vm-r4.2-current   3.1 KiB
 systemd-devel            x86_64 257.13-1.fc42 updates               612.1 KiB
Transaction Summary:
 Installing:         3 packages
Total size of inbound packages is 746 KiB. Need to download 746 KiB.
After this operation, 835 KiB extra will be used (install 835 KiB, remove 0 B).
Is this ok [y/N]: Operation aborted by the user.
+ :
```
Cancelled ^^^

run-tests
marmarek, line 1: I don't like this change, when user calls the script directly on their own system, it should at least ask for confirmation.
The failure you got in CI was because you forgot to add the package to CI config (.gitlab-ci.yml).

marmarek (changes requested):

marmarek (approved):

PR #239, opened 2026-07-24
qrexec-agent: avoid blocking on incomplete trigger requests

### LINKED ISSUE #11010: qrexec-agent: an incomplete local trigger request blocks the main event loop
## Qubes OS release

Qubes OS 4.3

## Brief summary

`qrexec-agent` handles a newly accepted local trigger socket synchronously in its main event loop. If a local client keeps the connection open without completing the request framing, `handle_trigger_io()` remains in `read_all()` and subsequent local qrexec trigger requests are not processed until that connection is closed.

I confirmed this with a bounded, three-second test in a dedicated disposable test AppVM. This appears to be a local availability/robustness issue in the source AppVM. I have not observed a policy bypass, cross-VM access, dom0 impact, privilege escalation, or data disclosure.

## Steps to reproduce

1. Start a disposable AppVM with the standard Qubes agent.
2. As the normal unprivileged VM user, open the local qrexec-agent trigger endpoint and keep one request intentionally incomplete for a short, fixed interval.
3. While that connection is open, issue a harmless qrexec request from the same VM with a short client-side timeout.
4. Observe that the second request times out while the incomplete connection remains open.
5. Close the incomplete connection and repeat the harmless request. It is processed normally again (in my test environment, it reached the existing policy denial path).

I am intentionally not publishing the raw trigger payload in this issue. I can provide the small bounded reproducer privately to maintainers if useful.

## Expected behavior

An incomplete or stalled local trigger client should not prevent `qrexec-agent` from continuing to service other local requests. The agent should use non-blocking framed reads, per-client state, or a bounded read deadline and return to its event loop.

## Actual behavior

The agent blocks in the first `read_all()` call for the accepted client. In my controlled test, a normal request made during the three-second window timed out with exit code `124`; after the incomplete connection was released, the same request returned the expected policy refusal (`126`).

## Additional information

Tested source: `qubes-core-qrexec` v4.3.12, commit `cc801b8`.

The same synchronous read pattern is currently present on the upstream `main` branch:

https://github.com/QubesOS/qubes-core-qrexec/blob/main/agent/qrexec-agent.c#L822-L862

The relevant sequence is:

```c
client_fd = do_accept(trigger_fd);
if (!read_all(client_fd, &hdr, sizeof(hdr)))
    goto error;
```

The `libvchan_send()` calls occur only after the complete header and request body have been read. This report concerns event-loop availability before any request is forwarded to dom0.

A regression test should keep one local trigger connection incomplete and verify that a second request remains serviceable or is rejected after a bounded timeout.

iasds:
## Summary

- accept local trigger clients without blocking the agent event loop
- track incomplete trigger requests and read their header and body incrementally via `ppoll`
- bound incomplete requests with a timeout and evict the oldest pending client at capacity
- add socket regression tests for a competing valid request and fragmented request framing

## Test plan

- Reproduced the original issue in an AppVM: baseline `0.01s`, blocked `3.01s` (exit 124), recovered `0.01s`.
- `gcc -fsyntax-only` for `agent/qrexec-agent.c` using the project include paths and a minimal `libvchan` API stub; the AppVM lacks the real `libvchan` development package.
- `python3 -m py_compile qrexec/tests/socket/agent.py`
- `git diff --check`

Fixes QubesOS/qubes-issues#11010

### DIFF
--- a/agent/qrexec-agent.c
+++ b/agent/qrexec-agent.c
@@ -852,7 +852,9 @@ static void expire_trigger_clients(void)
     for (size_t i = 0; i < MAX_FDS; i++) {
         struct trigger_client *client = &trigger_clients[i];
         if (client->fd != -1 &&
-            now.tv_sec - client->accepted_at.tv_sec >= TRIGGER_CLIENT_TIMEOUT) {
+            (now.tv_sec - client->accepted_at.tv_sec > TRIGGER_CLIENT_TIMEOUT ||
+             (now.tv_sec - client->accepted_at.tv_sec == TRIGGER_CLIENT_TIMEOUT &&
+              now.tv_nsec >= client->accepted_at.tv_nsec))) {
             LOG(WARNING, "Timed out waiting for trigger request");
             close_trigger_client(client);
         }
--- a/agent/qrexec-agent.c
+++ b/agent/qrexec-agent.c
@@ -38,6 +38,7 @@
 #include <sys/stat.h>
 #include <assert.h>
 #include <limits.h>
+#include <time.h>
 #ifdef HAVE_PAM
 #include <security/pam_appl.h>
 #endif
@@ -62,10 +63,21 @@ struct waiting_request {
     struct qrexec_parsed_command *cmd;
 };
 
+struct trigger_client {
+    int fd;
+    struct msg_header hdr;
+    size_t hdr_received;
+    struct trigger_service_params4 *params;
+    size_t params_received;
+    struct timespec accepted_at;
+};
+
 /*  */
 static struct connection_info connection_info[MAX_FDS];
 
 static struct waiting_request requests_waiting_for_session[MAX_FDS];
+
+static struct trigger_client trigger_clients[MAX_FDS];
 
 static libvchan_t *ctrl_vchan;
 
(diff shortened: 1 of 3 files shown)


agent/qrexec-agent.c
```diff
@@ -819,47 +835,143 @@ static void reap_children(void)
...
+
+static void handle_trigger_io(void)
+{
+    int client_fd = do_accept(trigger_fd);
+    if (client_fd < 0)
+        return;
+
+    int flags = fcntl(client_fd, F_GETFL, 0);
+    if (flags < 0 || fcntl(client_fd, F_SETFL, flags | O_NONBLOCK) < 0) {
+        close(client_fd);
```
marmarek, line 945: Log error here.

agent/qrexec-agent.c
```diff
@@ -819,47 +835,143 @@ static void reap_children(void)
...
-    params = malloc(hdr.len);
-    if (!params)
-        goto error;
-    if (!read_all(client_fd, params, hdr.len))
+    if (!client->params) {
+        client->params = malloc(client->hdr.len);
+        if (!client->params)
+            goto error;
+    }
+    ret = read_trigger_client(client, client->params, &client->params_received,
```
marmarek, line 910: If reads stop here (this one returns 0), next iteration will read the data into the header buffer again. I think you may use `client->params == NULL` as a condition for reading the header above.

marmarek (changes requested):

agent/qrexec-agent.c
marmarek: Actually, it won't really read anything into the header buffer, it will exit early because `client->hdr_received == sizeof(client->hdr)` already. But it could use at least a comment if not explicit condition in `handle_trigger_client_io()` to make it clearer.

